PR #169, opened 2020-12-07
Added `pino` logging library in the backend

ferdingler:
## Description

Added a library to do logging as well as an environment variable to control the logging level or disable it entirely in the Lambda functions. I also added `logger.debug()` in the DynamoDB Service just as an example of how to use the logger. 

## Testing

Deployed to my personal environment and verified it works. I also verified that CloudWatch Logs Insights is able to parse the logs in JSON format and identify the different fields in the log so that you can use those fields in your queries. 

---


trietlu (approved):
lgtm.  Thanks for validating with CloudWatch Insights.  It's also compatible with Elastic based on docs.

PR #170, opened 2020-12-09
GTT-742 Using React Context to create a SettingsProvider

ferdingler:
## Description

Used React Context to create a `SettingsProvider` that fetches the Settings from the backend only once when the application loads. All components in the Tree can consume these settings without having to refetch them. 

**Note**: This doesn't include fetching settings for Public users. We need a separate endpoint for that and we need to handle both cases accordingly in the `SettingsProvider`. 

## Testing

I had to update a few unit tests because `loadingSettings` is no longer a variable that the `useSettings` hook returns. I also tested thoroughly in my personal environment. 

---

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { BrowserRouter as Router, Switch, Route } from "react-router-dom";
 import { withAuthenticator } from "@aws-amplify/ui-react";
+import SettingsProvider from "./context/SettingsProvider";
 
 import withMainLayout from "./layouts/Main";
 import withAdminLayout from "./layouts/Admin";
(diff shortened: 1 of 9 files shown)


trietlu:
@ferdingler What happens if another user updates a field in Settings (like publishing guidance)? We wouldn't know to reload right?

ferdingler:
> @ferdingler What happens if another user updates a field in Settings (like publishing guidance)? We wouldn't know to reload right?

hmm good point. The backend would prevent the old data to overwrite the new one because of the `updatedAt` field. We do a ConditionalWrite in DynamoDB to make sure the new data has the latest `updatedAt`. However, we should probably reload the Settings in the Edit Publishing Guidance screen anyways. I can add that in a separate PR if that's ok with you. I would like to test a few different ways of doing so.

miguelpdiaz8 (approved):
It's ok to add it in a different PR.

PR #177, opened 2020-12-11
GTT-761 Bug Fixes and refactored layouts

ferdingler:
## Description

- GTT-755: Bug fix for the Logout button in the menu being black on a black background. 
- Refactored the Footer to make it a normal component, not a High Order Component. 
- Changed normal <a> links to make them proper links with `<Link>` in the layouts. 
- Renamed `Main` layout for `Public` layout instead. 
- Created a new route for 404 so we can redirect to that page from other screens by using the `<Redirect>` component from react router. 

## Testing

You can test on my personal environment. Check that the Logout button shows in white text color: https://fdingler.badger.wwps.aws.dev/admin/dashboards. 

You can also test the 404 redirection works with an invalid dashboard: 
https://fdingler.badger.wwps.aws.dev/123456789

---

### DIFF
--- a/frontend/src/components/MarkdownRender.css
+++ b/frontend/src/components/MarkdownRender.css
@@ -1,15 +1,15 @@
-a {
+.Markdown a {
   color: #565c65 !important;
 }
 
-a:hover {
+.Markdown a:hover {
   color: #1b1b1b !important;
 }
 
-a:active {
+.Markdown a:active {
   color: #1b1b1b !important;
 }
 
-a:visited {
+.Markdown a:visited {
   color: #1b1b1b !important;
 }
--- a/frontend/src/containers/ViewDashboard.tsx
+++ b/frontend/src/containers/ViewDashboard.tsx
@@ -1,12 +1,10 @@
 import React from "react";
-import { useParams } from "react-router-dom";
+import { useParams, Redirect } from "react-router-dom";
 import Link from "../components/Link";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faArrowLeft } from "@fortawesome/free-solid-svg-icons";
 import { usePublicDashboard } from "../hooks";
 import WidgetRender from "../components/WidgetRender";
-import FourZeroFour from "./FourZeroFour";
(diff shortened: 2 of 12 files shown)


frontend/src/components/MarkdownRender.css
```diff
@@ -1,15 +1,15 @@
...
 
-a:active {
+.Markdown a:active {
   color: #1b1b1b !important;
 }
 
-a:visited {
+.Markdown a:visited {
   color: #1b1b1b !important;
 }
```
ferdingler, line 15: this was the cause for the GTT-755 bug. These styles were being applied to all the website. I added this Markdown class so that they only apply within the markdown component.

frontend/src/containers/ViewDashboard.tsx
```diff
@@ -47,9 +45,7 @@ function ViewDashboard() {
       })}
     </>
   ) : (
-    <FooterLayout>
-      <FourZeroFour />
-    </FooterLayout>
+    <Redirect to="/404/page-not-found" />
```
miguelpdiaz8, line 48: Now, when we redirect in the public website, the FourZeroFour component does not have footer:

https://fdingler.badger.wwps.aws.dev/404/page-not-found

This can be fixed changing the component to withPublicLayout(FourZeroFour) in the 404/page-not-found route

And at the end of the switch in App.tsx add: ```<Redirect from='*' to='404/page-not-found' />```

miguelpdiaz8 (changes requested):

frontend/src/containers/ViewDashboard.tsx
miguelpdiaz8: The Redirect part in App.tsx is for completeness but only changing this: withPublicLayout(FourZeroFour) will fix the issue.

ferdingler: Does the footer apply only to Admin screens? Or also to Public screens?

miguelpdiaz8: Both

miguelpdiaz8 (approved):

PR #179, opened 2020-12-14
Data ingestion API

miguelpdiaz8:
## Description

Data ingestion API

## Testing

Added unit tests and tested it with Postman. 

---

### DIFF
--- a/backend/src/lib/factories/dataset-factory.ts
+++ b/backend/src/lib/factories/dataset-factory.ts
@@ -33,7 +33,9 @@ function fromItem(item: DatasetItem): Dataset {
     fileName: item.fileName,
     createdBy: item.createdBy,
     s3Key: item.s3Key,
-    sourceType: item.sourceType as SourceType,
+    sourceType: item.sourceType
+      ? (item.sourceType as SourceType)
+      : SourceType.FileUpload,
   };
   return dataset;
 }
--- a/backend/src/lib/factories/dataset-factory.ts
+++ b/backend/src/lib/factories/dataset-factory.ts
@@ -1,5 +1,5 @@
 import { v4 as uuidv4 } from "uuid";
-import { Dataset, DatasetItem } from "../models/dataset";
+import { Dataset, DatasetItem, SourceType } from "../models/dataset";
 
 const DATASET_ITEM_TYPE = "Dataset";
 
@@ -10,6 +10,7 @@ type DatasetInfo = {
     raw: string;
     json: string;
   };
+  sourceType: SourceType;
 };
 
 function createNew(info: DatasetInfo): Dataset {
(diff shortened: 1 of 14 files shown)


trietlu (commented):
Had a side discussion with @miguelpdiaz8 about using a different name instead of /ingestapi.  Having the "api" in the name of the resource is not ideal

backend/src/lib/factories/dataset-factory.ts
```diff
@@ -21,21 +22,41 @@ function createNew(info: DatasetInfo): Dataset {
...
 }
 
+function fromItem(item: DatasetItem): Dataset {
+  const id = item.pk.substring(8);
+  let dataset: Dataset = {
+    id,
+    fileName: item.fileName,
+    createdBy: item.createdBy,
+    s3Key: item.s3Key,
+    sourceType: item.sourceType as SourceType,
```
ferdingler: what about existing datasets that don't have a `sourceType` defined? We should probably default to `SourceType.fileUpload` here.

backend/src/lib/repositories/dataset-repo.ts
```diff
@@ -51,6 +86,90 @@ class DatasetRepository extends BaseRepository {
...
+    let fileExists = await this.s3Service.objectExists(
+      this.bucketName,
+      jsonKey
+    );
+    if (!fileExists) {
+      console.error("JSON file not found for dataset=", dataset);
+      throw new Error("JSON file for dataset not found on S3");
+    }
+
+    this.dynamodb.put({
```
ferdingler: missing `await` here?

backend/src/lib/repositories/dataset-repo.ts
```diff
@@ -51,6 +86,90 @@ class DatasetRepository extends BaseRepository {
...
+  public async updateDataset(id: string, metadata: any, data: any) {
+    const dataset = await this.getDatasetById(id);
+    const jsonS3Key = dataset.s3Key.json || this.getNewJsonS3Key();
+    const jsonKey = this.s3Prefix.concat(jsonS3Key);
+    await this.s3Service.putDataset(this.bucketName, jsonKey, data);
+
+    let fileExists = await this.s3Service.objectExists(
+      this.bucketName,
+      jsonKey
+    );
```
ferdingler: Is this necessary? If you just uploaded the object on line 122 why do you need to check if it exists?

backend/src/lib/services/s3.ts
```diff
@@ -19,6 +20,17 @@ class S3Service {
     return S3Service.instance;
   }
 
+  async putDataset(bucketName: string, jsonS3Key: string, jsonFile: string) {
```
ferdingler: I think the `S3Service` shouldn't have knowledge of the concept of `Dataset`. I would suggest to rename this function to `putObject` instead so that it is generic for any type of file.

backend/src/lib/services/s3.ts
```diff
@@ -32,6 +44,14 @@ class S3Service {
       return false;
     }
   }
+
+  async deleteDataset(bucketName: string, jsonS3Key: string) {
```
ferdingler: Same comment as above. I would suggest to leave this function name generic `deleteObject`.

cdk/lib/constructs/api.ts
```diff
@@ -104,8 +104,20 @@ export class BackendApi extends cdk.Construct {
...
+    const dataset = datasets.addResource("{id}");
+    dataset.addMethod("GET", apiIntegration, methodProps);
+
+    const ingestApi = this.api.root.addResource("ingestapi");
+    const ingestApiDatasets = ingestApi.addResource("dataset");
+    ingestApiDatasets.addMethod("POST", apiIntegration, methodProps);
+
+    const ingestApiDataset = ingestApiDatasets.addResource("{id}");
+    ingestApiDataset.addMethod("PUT", apiIntegration, methodProps);
+    ingestApiDataset.addMethod("DELETE", apiIntegration, methodProps);
```
ferdingler, line 119: This is where Authentication comes into place. This is fine for now, but we should change this to `API KEY` authentication in another Pull Request.

ferdingler (changes requested):

ferdingler (approved):

PR #184, opened 2020-12-16
Secure data ingestion API

miguelpdiaz8:
## Description

Secure data ingestion API

## Testing

Tested the endpoints with the API Key using Postman

---

### DIFF
--- a/cdk/lib/constructs/api.ts
+++ b/cdk/lib/constructs/api.ts
@@ -44,6 +44,19 @@ export class BackendApi extends cdk.Construct {
             },
           },
         }),
+
+        //Need to change this policy for ingest API with specific values
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: ["execute-api:Invoke"],
+          resources: ["execute-api:/prod/*/ingestapi/*"],
+          principals: [new AnyPrincipal()],
+          conditions: {
+            IpAddress: {
+              "aws:SourceIp": ["0.0.0.0/0"],
+            },
+          },
+        }),
       ],
     });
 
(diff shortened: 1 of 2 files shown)


cdk/lib/constructs/api.ts
```diff
@@ -25,13 +31,30 @@ export class BackendApi extends cdk.Construct {
       }
     );
 
+    const apiPolicy = new PolicyDocument({
+      statements: [
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: ["execute-api:Invoke"],
+          resources: ["execute-api:/prod/*/*"],
```
ferdingler, line 39: This policy should only apply to `execute-api:/prod/ingestapi/*` right? We don't want to apply this same Resource Policy to the other apis.

cdk/lib/constructs/api.ts
```diff
@@ -44,6 +67,21 @@ export class BackendApi extends cdk.Construct {
 
     this.addPrivateEndpoints(apiIntegration, authorizer);
     this.addPublicEndpoints(publicApiIntegration);
+
+    const key = this.api.addApiKey("PerformanceDashboardApiKey");
```
ferdingler: I would call this key `IngestApiKey` or `PerfDashIngestApiKey` because it's really just for the ingest api. It won't work for the other apis.

cdk/lib/constructs/api.ts
```diff
@@ -44,6 +67,21 @@ export class BackendApi extends cdk.Construct {
...
+    const plan = this.api.addUsagePlan("PerformanceDashboardUsagePlan", {
+      name: "PerformanceDashboardUsagePlan",
+      apiKey: key,
+      throttle: {
+        rateLimit: 100,
+        burstLimit: 200,
```
ferdingler: I suggest we go a little bit more conservative here. 100/200 requests per second is a lot considering that an AWS account has 1,000 concurrent lambda executions by default and considering that the Ingest API has to load data into S3 so the latency is not as good as it is for other endpoints. 

My suggestion is: 
```
rateLimit: 25
burstLimit: 50
```

cdk/lib/constructs/api.ts
```diff
@@ -25,13 +31,43 @@ export class BackendApi extends cdk.Construct {
...
+          effect: Effect.ALLOW,
+          actions: ["execute-api:Invoke"],
+          resources: ["execute-api:/prod/*/*"],
+          principals: [new AnyPrincipal()],
+          conditions: {
+            IpAddress: {
+              "aws:SourceIp": ["0.0.0.0/0"],
+            },
+          },
+        }),
```
ferdingler, line 46: Do we need to have a ResourcePolicy here? Isn't the one below (Line 49) already covering what we need?

miguelpdiaz8: We need it, if not we are only allowing ingestapi/ We need to give access to the rest if we add a resource policy.

ferdingler: ok

ferdingler (approved):

PR #198, opened 2020-12-22
Backend: Check if dataset is being used by a Widget before deleting (return 409 error if so)

miguelpdiaz8:
## Description

Backend: Check if dataset is being used by a Widget before deleting (return 409 error if so)

## Testing

Tested it in my local environment with Postman.

---


backend/src/lib/repositories/dataset-repo.ts
```diff
@@ -200,6 +202,34 @@ class DatasetRepository extends BaseRepository {
...
+
+    if (!result.Items) {
+      return 0;
+    }
+
+    const items = result.Items.filter(
+      (item) =>
+        item.widgetType !== WidgetType.Text &&
+        item.content.s3Key.json === dataset.s3Key.json
+    );
```
ferdingler: You should add a [FilterExpression](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.QueryFilter.html) to the DynamoDB query itself so that we don't have to fetch all widgets and then filter them programmatically. 

Also, this query is likely to need pagination. The number of Widgets across all dashboards across all versions is potentially large such that you may need additional pagination requests to fetch them all. Alternatively, you can add a GSI to the table to fetch Widgets by DatasetId and then you don't need any filters and we likely won't need pagination either.

ferdingler (dismissed):

backend/src/lib/repositories/dataset-repo.ts
trietlu: GSI is a good idea.  The only thing about GSI is that it's not real time.  So a widget and dataset could be added, and before the GSI is updated,  a /delete is called and the widget count might return 0

ferdingler: hmm good point. 🤔

miguelpdiaz8: We are already using another GSI in the current solution and similar places in the code, this is something we need to fix, but since it is very short delay, we can use a GSI and think about better approach later for all cases.

miguelpdiaz8: The issue is making a global secondary index in a field that is a Map, that is not allowed.

ferdingler: But the GSI should be on the 'datasetId' field though, which is a String, not a map.

ferdingler: I think the misunderstanding is that we should delete a dataset by its ID not by the JSON key.

miguelpdiaz8: Yes, but anyway the datasetId is inside a json object(Map) in dynamo

miguelpdiaz8: DatasetId is inside the content object, does not exist outside. I'm doing it now with pagination

trietlu (approved):

PR #203, opened 2020-12-29
Add navbarTitle to the Settings object

jenboydclark:
## Description

Added navbarTitle to the Settings object

## Testing

https://dvinr4n9hngnv.cloudfront.net

---

### DIFF
--- a/backend/src/lib/controllers/settings-ctrl.ts
+++ b/backend/src/lib/controllers/settings-ctrl.ts
@@ -26,7 +26,7 @@ async function updateSettings(req: Request, res: Response) {
   }
 
   let { updatedAt } = req.body;
-  const { publishingGuidance, dateTimeFormat } = req.body;
+  const { publishingGuidance, dateTimeFormat, navbarTitle } = req.body;
 
   if (!updatedAt) {
     res.status(400);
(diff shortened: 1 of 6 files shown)


backend/src/lib/models/settings.ts
```diff
@@ -24,4 +26,5 @@ export interface SettingsItem {
     date: string;
     time: string;
   };
+  navbarTitle: string;
```
ferdingler: Let's make this property optional here (like dateTimeFormat and publishingGuidance) because existing DynamoDB items won't have this property until the next time it gets saved.

backend/src/lib/controllers/settings-ctrl.ts
```diff
@@ -26,7 +26,7 @@ async function updateSettings(req: Request, res: Response) {
   }
 
   let { updatedAt } = req.body;
-  const { publishingGuidance, dateTimeFormat } = req.body;
+  const { publishingGuidance, dateTimeFormat, navbarTitle } = req.body;
```
ferdingler, line 29: Can you please also add a test in the `settings-ctrl.test.ts` for this scenario of updating the navbarTitle.

ferdingler (changes requested):

ferdingler (approved):
Thanks for making the changes!

PR #206, opened 2020-12-31
GTT-548 ShowTitle option for add/edit content

ghost:
Admin have the option to hide or show titles when creating or editing widgets. Titles can be hidden or shown during preview and that setting will be saved for the published dashboard. 

- Checkbox element added to chart, text, and table containers 
- Conditional statements added to chart, text, table containers and widgets to hide or show the title
- widget model updated to accept showTitle parameter
-  widget functions updated to accept showTitle parameter

### DIFF
--- a/frontend/src/components/TextWidget.tsx
+++ b/frontend/src/components/TextWidget.tsx
@@ -7,11 +7,11 @@ interface Props {
 }
 
 function TextWidget(props: Props) {
-  const { content } = props.widget;
+  const { content, showTitle, name } = props.widget;
 
   return (
     <div className="margin-left-1">
-      {props.widget.showTitle ? <h2>{props.widget.name}</h2> : ""}
+      {showTitle && <h2>{name}</h2>}
       <MarkdownRender source={content.text} />
     </div>
   );
--- a/frontend/src/components/TextWidget.tsx
+++ b/frontend/src/components/TextWidget.tsx
@@ -11,7 +11,7 @@ function TextWidget(props: Props) {
 
   return (
     <div className="margin-left-1">
-      <h2>{props.widget.name}</h2>
+      {props.widget.showTitle ? <h2>{props.widget.name}</h2> : ""}
       <MarkdownRender source={content.text} />
     </div>
   );
--- a/frontend/src/containers/AddChart.tsx
+++ b/frontend/src/containers/AddChart.tsx
@@ -85,7 +85,6 @@ function AddChart() {
   };
 
   const onSubmit = async (values: FormValues) => {
-    return;
     try {
       let newDataset;
       if (csvFile) {
--- a/frontend/src/containers/AddChart.tsx
+++ b/frontend/src/containers/AddChart.tsx
@@ -25,6 +25,7 @@ interface FormValues {
   title: string;
   summary: string;
   chartType: string;
+  showTitle: boolean;
 }
 
 interface PathParams {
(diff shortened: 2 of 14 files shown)


frontend/src/components/TextWidget.tsx
```diff
@@ -11,7 +11,7 @@ function TextWidget(props: Props) {
 
   return (
     <div className="margin-left-1">
-      <h2>{props.widget.name}</h2>
+      {props.widget.showTitle ? <h2>{props.widget.name}</h2> : ""}
```
ferdingler: An alternative shorter syntax would be: 
```js
{props.widget.showTitle && (<h2>{props.widget.name}</h2>)}
```

frontend/src/containers/AddChart.tsx
```diff
@@ -55,6 +56,7 @@ function AddChart() {
   const [title, setTitle] = useState("");
   const [summary, setSummary] = useState("");
   const [chartType, setChartType] = useState<ChartType>(ChartType.LineChart);
+  const [showTitle, setShowTitle] = useState(true);
```
ferdingler, line 59: Instead of having this state variable and the `handleShowTitleChange` function. You can simply get the value of the checkbox by using the `getValues()` function that you get from the `useForm<FormValues>()` hook in Line 41. .

ferdingler: So, in Line 530 for example, you can do something like this instead: 

```
<LineChartPreview
  title={getValues().showTitle ? title : ""}
/>
```

frontend/src/containers/AddChart.tsx
```diff
@@ -280,6 +287,21 @@ function AddChart() {
...
               />
 
+              <div className="usa-checkbox">
+                <input
+                  className="usa-checkbox__input"
+                  id="display-title"
+                  type="checkbox"
+                  name="showTitle"
+                  defaultChecked={true}
+                  onChange={handleShowTitleChange}
```
ferdingler, line 297: This onChange won't be necessary if you address my comments above.

frontend/src/containers/AddText.tsx
```diff
@@ -28,6 +29,7 @@ function AddText() {
   const [creatingWidget, setCreatingWidget] = useState(false);
   const [title, setTitle] = useState("");
   const [text, setText] = useState("");
+  const [showTitle, setShowTitle] = useState(true);
```
ferdingler, line 32: Same here. No need to have `showTitle` in the state. Just get it from `getValues()` when you need it.

frontend/src/containers/AddText.tsx
```diff
@@ -153,7 +171,11 @@ function AddText() {
         </div>
         <div className="grid-col-6">
           <h4 className="margin-top-4">Preview</h4>
-          <h2 className="margin-top-4 margin-left-2px">{title}</h2>
+          {showTitle ? (
+            <h2 className="margin-top-4 margin-left-2px">{title}</h2>
+          ) : (
+            ""
+          )}
```
ferdingler, line 178: Same as other comment. Shorter syntax would be: 
```
{showTitle && (<h2 className="margin-top-4 margin-left-2px">{title}</h2>)}
```

ferdingler (changes requested):
FYI, looks like the unit tests are failing too.

frontend/src/containers/AddChart.tsx
ghost: For some reason, implementing this change results in the preview not updating appropriately. The preview doesn't update when the checkbox is toggled, only when the summary or title is being edited.

ferdingler: I see. Ok that's fine then.

ferdingler (approved):

PR #209, opened 2021-01-06
Added -optional label to the add/edit table screens

ajmokotoff:
## Description

We wanted to make it more clear to our users that the table summary input field is optional and not required. Every other input on the edit/add page is required. So, I simply added "- optional" to the label for the input, on both the edit table page, and the add table page.

## Testing

I made the changes locally, and then ran the test suite (./test.sh). I also double checked the webpage it self, to ensure the -optional label was there. Both tests completed successfully.

---


miguelpdiaz8 (approved):

ferdingler (approved):
Nicely done

PR #212, opened 2021-01-06
BugFix: Content items overlap on dashboard

miguelpdiaz8:
## Description

Content items overlap on dashboard

## Testing

---

### DIFF
--- a/frontend/src/components/PartWholeChartPreview.tsx
+++ b/frontend/src/components/PartWholeChartPreview.tsx
@@ -54,18 +54,16 @@ const PartWholeChartPreview = (props: Props) => {
   };
 
   return (
-    <div
-      style={{
-        width: "100%",
-        height: `${props.data && props.data.length > 15 ? "600px" : "300px"}`,
-      }}
-    >
+    <div>
       <h2 className="margin-left-1 margin-bottom-1">{props.title}</h2>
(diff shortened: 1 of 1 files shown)


frontend/src/components/PartWholeChartPreview.tsx
```diff
@@ -54,18 +54,16 @@ const PartWholeChartPreview = (props: Props) => {
...
+    <div>
       <h2 className="margin-left-1 margin-bottom-1">{props.title}</h2>
       <p className="margin-left-1 margin-top-0 margin-bottom-3">
         {props.summary}
       </p>
       {partWholeData.length && (
-        <ResponsiveContainer width="100%" height="100%">
+        <ResponsiveContainer
+          width="100%"
+          height={props.data && props.data.length > 15 ? 600 : 300}
```
ferdingler, line 65: Does this always work? What happens if the chart height goes beyond 600px? Will it overlap again?

miguelpdiaz8: Yes I was thinking to do a similar approach that works for all sizes but this will work until 30 more or less, more than that currently is not working anyway. We can discuss offline, by instance if you try to create a chart now with covid vaccine dataset is not working.

ferdingler (approved):

PR #215, opened 2021-01-07
Backend - New endpoint to list users

miguelpdiaz8:
## Description

Backend - New endpoint to list users

## Testing

Tested it using Postman here: https://tzvlbpxa33.execute-api.us-east-1.amazonaws.com/prod/user

---


backend/src/lib/factories/user-factory.ts
```diff
@@ -0,0 +1,25 @@
+import { User } from "../models/user";
+import { UserType } from "aws-sdk/clients/cognitoidentityserviceprovider";
+
+function toUser(cognitoUser: UserType): User {
```
ferdingler: I think this function should be named `fromUserType` or `fromCognitoUser` because it creates a `User` from a Cognito `UserType` object. Similar to how we have `fromItem` in the other factories to convert raw dynamodb objects into our Domain Models.

backend/src/lib/repositories/__tests__/user-repo.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        ? cognitoUser.Attributes[1].Value === "true"
+        : false,
+      email: cognitoUser.Attributes ? cognitoUser.Attributes[2].Value : "",
+      createdAt: cognitoUser.UserCreateDate
+        ? cognitoUser.UserCreateDate
+        : new Date(),
+      updatedAt: cognitoUser.UserLastModifiedDate
+        ? cognitoUser.UserLastModifiedDate
+        : new Date(),
+    });
```
ferdingler: You should be expecting actual values and not have if conditions like `cognitoUser.Username || ""` in your assertions. 

Should be something like: 
```JS
expect(list[0]).toEqual({'
  userId: "test user",
  enabled: true,
  userStatus: "CONFIRMED",
  ...
})
```

backend/src/lib/factories/user-factory.ts
```diff
@@ -0,0 +1,25 @@
...
+  return {
+    userId: cognitoUser.Username || "",
+    enabled: cognitoUser.Enabled,
+    userStatus: cognitoUser.UserStatus,
+    sub: cognitoUser.Attributes ? cognitoUser.Attributes[0].Value : "",
+    emailVerified: cognitoUser.Attributes
+      ? cognitoUser.Attributes[1].Value === "true"
```
ferdingler: Is this attribute always in the position `1` of the Attributes array? I'm not sure if we should rely on that index. Also, we don't need to know if the email is verified. I would suggest you can get rid of this field.

backend/src/lib/factories/user-factory.ts
```diff
@@ -0,0 +1,25 @@
...
+      : false,
+    email: cognitoUser.Attributes ? cognitoUser.Attributes[2].Value : "",
```
ferdingler: Same. Relying on the index position of the Attributes array seems suspicious.

cdk/bin/main.ts
```diff
@@ -27,7 +27,7 @@ const auth = new AuthStack(app, "Auth", {
 
 const backend = new BackendStack(app, "Backend", {
   stackName: stackPrefix.concat("-Backend"),
-  userPoolArn: auth.userPoolArn,
+  auth,
```
ferdingler: I'm not sure if it's a good idea to pass the entire stack. It creates tighter coupling than there should be. Why not passing just the UserPool object?

ferdingler (changes requested):

cdk/bin/main.ts
miguelpdiaz8: We dont have an userPool object, the alternative is to pass:

`userPool: {
    id: auth.userPoolId,
    arn: auth.userPoolArn,
}`

or the two separate strings

ferdingler (approved):
Thanks for making the changes!

PR #223, opened 2021-01-11
Backend - New endpoint to create users in bulk

miguelpdiaz8:
## Description

Backend - New endpoint to create users in bulk

## Testing

Tested it using postman and added unit tests.

---

### DIFF
--- a/backend/src/lib/controllers/user-ctrl.ts
+++ b/backend/src/lib/controllers/user-ctrl.ts
@@ -28,7 +28,18 @@ async function addUsers(req: Request, res: Response) {
     return res.send("Unauthorized");
   }
 
+  const { role } = req.body;
   const { emails } = req.query;
+
+  if (!role) {
+    res.status(400).send("Missing required body `role`");
+    return;
+  }
+
+  if (role !== "Admin" && role !== "Editor" && role !== "Publisher") {
+    res.status(400).send("Invalid role value");
+    return;
+  }
 
   if (!emails) {
     res.status(400).send("Missing required query `emails`");
@@ -47,7 +58,7 @@ async function addUsers(req: Request, res: Response) {
   const repo = UserRepository.getInstance();
   try {
     await repo.addUsers(
-      userEmails.map((email) => UserFactory.createNew(email))
+      userEmails.map((email) => UserFactory.createNew(email, role))
     );
     return res.send();
   } catch (error) {
--- a/backend/src/lib/controllers/user-ctrl.ts
+++ b/backend/src/lib/controllers/user-ctrl.ts
@@ -1,5 +1,6 @@
 import { Request, Response } from "express";
 import UserRepository from "../repositories/user-repo";
+import UserFactory from "../factories/user-factory";
 import AuthService from "../services/auth";
 
 async function getUsers(req: Request, res: Response) {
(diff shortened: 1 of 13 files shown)


backend/src/lib/controllers/user-ctrl.ts
```diff
@@ -19,6 +20,57 @@ async function getUsers(req: Request, res: Response) {
...
+async function addUsers(req: Request, res: Response) {
+  const user = AuthService.getCurrentUser(req);
+
+  if (!user) {
+    res.status(401);
+    return res.send("Unauthorized");
+  }
+
+  const { role } = req.body;
+  const { emails } = req.query;
```
ferdingler: Why are the email addresses in the query params and not in the body? This might be a security concern because query parameters get saved in the browsers history. 

Ref: https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url

backend/src/lib/controllers/user-ctrl.ts
```diff
@@ -19,6 +20,57 @@ async function getUsers(req: Request, res: Response) {
...
+
+  if (!role) {
+    res.status(400).send("Missing required body `role`");
+    return;
+  }
+
+  if (role !== "Admin" && role !== "Editor" && role !== "Publisher") {
```
ferdingler: Instead of hardcoding the strings here, you could use the `enum Role` you created. 

```
if (role !== Roles.Admin ...)
```

backend/src/lib/controllers/user-ctrl.ts
```diff
@@ -19,6 +20,57 @@ async function getUsers(req: Request, res: Response) {
...
+
+  if (!emails) {
+    res.status(400).send("Missing required query `emails`");
+    return;
+  }
+
+  const userEmails = (emails as string).split(",");
+
+  for (const userEmail of userEmails) {
+    if (!emailIsValid(userEmail)) {
```
ferdingler, line 52: Nice check!

cdk/lib/auth-stack.ts
```diff
@@ -48,6 +50,17 @@ export class AuthStack extends cdk.Stack {
...
+    if (process.env.COGNITO_EMAIL) {
+      new cognito.CfnUserPoolUser(this, "UserPoolUser", {
+        userPoolId: pool.userPoolId,
+        username: process.env.COGNITO_EMAIL.split("@")[0],
+        userAttributes: [
+          { name: "email", value: process.env.COGNITO_EMAIL },
+          { name: "custom:roles", value: JSON.stringify(["Admin"]) },
+        ],
+      });
+    }
```
ferdingler: This approach works well with CDK deployments, but we need to make this work with CloudFormation parameters. Customers will be deploying Badger using a CloudFormation template and environment variables is not something they can provide. 

We can do something like this: 
```
const adminEmail = new CfnParameter(this, "adminEmail");

if (adminEmail) {
  new cognito.CfnUserPoolUser(this, "UserPoolUser", {
          userPoolId: pool.userPoolId,
          username: adminEmail,
          userAttributes: [
            { name: "email", value: adminEmail },
            { name: "custom:roles", value: JSON.stringify(["Admin"]) },
          ],
  });
}
```

ferdingler (approved):

PR #230, opened 2021-01-13
GTT-846: [Front-End] Modify Settings Page to Allow Edits to Topic Area Labels

ajmokotoff:
## Description

Added the front end changes to allow for the label "Topic Area" to be modified across the site.

## Testing

I added tests for the items added to the front end, I updated the snapshots, and confirmed that all other tests passed. I verified myself, by using testing various scenarios on the site itself.

---


frontend/src/containers/__tests__/TopicareaSettings.test.tsx
```diff
@@ -9,7 +9,18 @@ test("renders the title", async () => {
...
+    getByRole("heading", { name: "Topic areas (Ministries)" })
+  ).toBeInTheDocument();
+});
+
+test("renders the title with set value of topic areas", async () => {
+  const { getByRole } = render(<TopicareaSettings />, {
+    wrapper: MemoryRouter,
+  });
+  expect(
+    getByRole("heading", { name: "Topic areas (Ministries)" })
```
miguelpdiaz8: As we discussed in the standup we need to use here the topicAreaLabel we just set and not longer use the ()

frontend/src/models/index.tsx
```diff
@@ -152,6 +156,10 @@ export type Settings = {
   };
   updatedAt?: Date;
   navbarTitle?: string;
+  topicAreaLabels?: {
+    singular: string;
+    plural: string;
+  };
```
ferdingler: I don't think we should make this property optional. Remove the `?`.  I noticed that you added a default value in the `SetttingsProvider` file so we can assume that this property will always be present.

frontend/src/services/UtilsService.ts
```diff
@@ -38,9 +38,22 @@ function getChartTypeLabel(chartType: string): string {
...
+  currentTopicArea: string | undefined
+): string {
+  return `Topic areas ${
+    currentTopicArea &&
+    currentTopicArea.toLowerCase() === defaultTopicArea.toLowerCase()
+      ? ""
+      : ` (${currentTopicArea})`
+  }`;
+}
+
```
ferdingler: Nice solution for the title!

frontend/src/containers/__tests__/TopicareaSettings.test.tsx
ferdingler: That's true. I guess you no longer need the `determineTopicAreaString` function.

frontend/src/containers/TopicareaListing.tsx
```diff
@@ -153,7 +153,7 @@ function TopicareaListing() {
               <span>
                 <Button
                   onClick={createTopicArea}
-                >{`Create ${EnvConfig.topicAreaLabel.toLowerCase()}`}</Button>
+                >{`Create new ${settings.topicAreaLabels?.singular.toLowerCase()}`}</Button>
```
ferdingler: If we make the `topicAreaLabels` property not optional, we don't need these conditional `?` accessors here. I think it looks cleaner without them and it is safe to assume that the property will always be there given that it has default values in both the `SettingsProvider` and the backend endpoint.

ferdingler (changes requested):

frontend/src/containers/__tests__/TopicareaSettings.test.tsx
ajmokotoff: Done!

frontend/src/models/index.tsx
ajmokotoff: Agreed. Complete

frontend/src/services/UtilsService.ts
ajmokotoff: Removed, as we decided above

ferdingler (approved):

PR #232, opened 2021-01-13
Frontend - New button to resend invite

miguelpdiaz8:
## Description

Frontend - New button to resend invite

## Testing

Added unit tests and tested in my local environment.

---

### DIFF
--- a/frontend/src/components/Table.tsx
+++ b/frontend/src/components/Table.tsx
@@ -11,7 +11,7 @@ import {
 interface Props {
   selection: "multiple" | "none";
   initialSortByField?: string;
-  asc?: boolean;
+  initialSortAscending?: boolean;
   screenReaderField?: string;
   filterQuery?: string;
   className?: string;
@@ -47,10 +47,10 @@ function Table(props: Props) {
           () => [
             {
               id: props.initialSortByField || "",
-              desc: !props.asc,
+              desc: !props.initialSortAscending,
             },
           ],
-          [props.initialSortByField, props.asc]
+          [props.initialSortByField, props.initialSortAscending]
         ),
       },
     },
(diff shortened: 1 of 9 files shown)


frontend/src/components/Table.tsx
```diff
@@ -11,6 +11,7 @@ import {
 interface Props {
   selection: "multiple" | "none";
   initialSortByField?: string;
+  asc?: boolean;
```
ferdingler: Nit picking: To follow the same naming convention of the other Prop, can you call this property `initialSortAscending?: boolean`?

frontend/src/hooks/__mocks__/index.tsx
```diff
@@ -360,15 +360,18 @@ export function useDateTimeFormatter() {
...
 
 export function useUsers() {
+  const [users] = useState([
+    {
+      userId: "johndoe",
+      email: "johndoe@example.com",
+      status: "CONFIRMED",
+      roles: ["Admin"],
+    },
+  ]);
```
ferdingler, line 370: Nice. Preventing infinite loops ;)

frontend/src/containers/UserListing.tsx
```diff
@@ -55,8 +128,11 @@ function UserListing() {
             },
             {
               Header: "Role",
-              accessor: "",
-              Cell: () => "Admin",
+              accessor: "roles",
+              Cell: (props: any) =>
+                props.value && props.value.length
+                  ? (props.value as Array<UserRoles>).join(",")
+                  : "Admin",
```
ferdingler: I think we should display an empty value here instead of defaulting to `Admin`. Because defaulting to "Admin" gives a false perception that there is a role assigned, when it is not true.

ferdingler (changes requested):

ferdingler (approved):

PR #241, opened 2021-01-15
GTT-805 Chart/Table summary below option

jenboydclark:
## Description

Added the option to place the chart/table summary above or below the visualization

## Testing

https://dvinr4n9hngnv.cloudfront.net/

---


ferdingler (commented):
@jenboydclark existing Widgets in the database today will not have this property. So, in the frontend, `content.summaryBelow` will be undefined for all those widgets. Did you verify that you are handling this gracefully?

jenboydclark:
@ferdingler It works fine for my widgets that were written before this change. The code never tests for false explicitly so where the value is undefined it will be equivalent to false and the summary will be displayed above the visualization.

ferdingler (approved):

PR #246, opened 2021-01-21
GTT-907: Immer upgrade to fix vulnerability

ajmokotoff:
## Description

Upgraded immer to 8.0.1, as that verison is safe from the vulnerability the older versions had.

## Testing

Ran test suite successfully and played around the site briefly. Didn't see any issues.

---


miguelpdiaz8 (approved):

frontend/package.json
```diff
@@ -26,6 +26,7 @@
     "@types/recharts": "^1.8.13",
     "@types/uuid": "^8.0.0",
     "dayjs": "^1.8.36",
+    "immer": "8.0.1",
```
ferdingler: We don't need a direct dependency on this library so is not a good idea to add it to our package.json. Instead, we need to see what library we depend on uses it, and see if that library has already released a patch.

ferdingler (changes requested):

frontend/package.json
trietlu: react-dev-utils is being upgraded to immer 8.0.1.  Let's track this [issue](https://github.com/facebook/create-react-app/issues/10411)

ajmokotoff: Removed from package.json. But I guess we can wait and see if they fix it in react dev utils.

ferdingler:
What process did you follow to upgrade this? Did you edit the yarn.lock manually? 
 
Doing a `npm ls immer` in the frontend package, I see the following: 

```
performance-dashboard-frontend@1.0.0
└─┬ react-scripts@4.0.0
  └─┬ react-dev-utils@11.0.1
    └── immer@7.0.9
```

Immer is a dependency of react-dev-utils which is a dependency of react-scripts. I wonder if it's best to upgrade react-scripts assuming that they have already released a patch.

ferdingler:
If react-scripts hasn't released a patch, then I believe the proper way for upgrading a transitive dependency is to package.json resolutions: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

ajmokotoff:
> What process did you follow to upgrade this? Did you edit the yarn.lock manually?
> 
> Doing a `npm ls immer` in the frontend package, I see the following:
> 
> ```
> performance-dashboard-frontend@1.0.0
> └─┬ react-scripts@4.0.0
>   └─┬ react-dev-utils@11.0.1
>     └── immer@7.0.9
> ```
> 
> Immer is a dependency of react-dev-utils which is a dependency of react-scripts. I wonder if it's best to upgrade react-scripts assuming that they have already released a patch.

I ran the yarn upgrade on that script, I then changed the dependency in react dev utils, to depend on that new version. Maybe this is incorrect. I do think it would be better to just update the utils library itself.

ferdingler:
Closing as the solution will be to bump up `react-scripts` when they release a new version.

PR #251, opened 2021-01-22
GTT-881, GTT-882, GTT-883

ajmokotoff:
## Description

This pull request encompasses three tasks. GTT-881, GTT-882, and GTT-883. These tasks are to create a component that acts as an image preview for an uploaded image, a new function in storage service that uploads images using amplify to S3 buckets, and new screens to add image content items to a dashboard.

### Implementation explanation

For StorageService.ts, I decided to extract code out of uploadDataset, because it was uploading a file and a JSON. In my particular case, we wanted to upload a file. So by extracting those 2 services into their own functions, my uploadImage function is able to simply call the uploadFile function. I thought this made it more flexible.

I also decided to export the imageFileTypes, so that the AddImage component could have dynamic accepted file types. If we ever want to support different image types in the future, this will allow us to only change it in one place. You will see this in line 39 on AddImage.tsx. 

However, for line 189 in AddImage.tsx, I decided to leave this static. Since in our queue we are assigned to make this translatable, I was thinking it would be a lot more difficult to translate a dynamic string, as it would have to translate "Must be a", instead of the entire string. Happy to change this if someone disagrees.

## Testing

1. I started by adding tests for each module that I wrote and worked on. Ran all of these successfully.
2. I updated the snapshots of the new pages I added.
3. To test the storage service, I uploaded actual images and verified that they ended up in S3 bucket, by going to the AWS console and looking in the S3 bucket directly. I also added more tests to verify that I did not break anything.
4. I then played around with the front end and verified the behavior looked right and matched the wireframe.
5. To test the image preview module, I tried uploading massive images and tiny images. I verified that the massive ones were scaled down to fit in the div, and that tiny ones were aligned with the top left corner.

Below is the tiny image
Screen Shot 2021-01-22 at 16 20 35

Below is the massive image
Screen Shot 2021-01-22 at 16 20 46

---


frontend/src/components/__tests__/ImagePreview.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      summaryBelow={false}
+    />,
+    { wrapper: MemoryRouter }
+  );
+  expect(getByText("test title")).toBeInTheDocument();
+  expect(getByText("test summary")).toBeInTheDocument();
+  expect(getByText("test summary").nextSibling).toContainHTML("img");
+});
+
+test("renders the image preview component with the summary below the chart", async () => {
```
ferdingler, line 29: 👏

ferdingler: I learned today about nextSibling and previousSibling

frontend/src/containers/AddImage.tsx
```diff
@@ -0,0 +1,291 @@
...
+      const s3Key = await StorageService.uploadImage(imageFile);
+
+      await BackendService.createWidget(
+        dashboardId,
+        values.title,
+        WidgetType.Image,
+        values.showTitle,
+        {
+          title: title,
+          s3Key: s3Key,
```
ferdingler: Does this work? –– If I remember correctly from @jenboydclark's pull request. The `s3Key` property of an ImageWidget is an object like this: 

```
s3Key: {
  raw: "abc.png"
}
```

ferdingler:
Looks really good @ajmokotoff, nice job! Great unit tests too. In the future, I would recommend to keep Pull Requests as small as possible. This one wasn't easy to split into multiple PRs so I understand, but my rule of thumb is if a PR starts approaching the +500 lines of code, is probably too big and ideally should be broken up. The larger a PR, the more likely reviewers will miss stuff. So, small PRs are always a better practice. 

I added a comment/question about the `s3Key` attribute. I believe you need to add a nested `raw` property just like the other widgets have it.

ajmokotoff:
> Looks really good @ajmokotoff, nice job! Great unit tests too. In the future, I would recommend to keep Pull Requests as small as possible. This one wasn't easy to split into multiple PRs so I understand, but my rule of thumb is if a PR starts approaching the +500 lines of code, is probably too big and ideally should be broken up. The larger a PR, the more likely reviewers will miss stuff. So, small PRs are always a better practice.
> 
> I added a comment/question about the `s3Key` attribute. I believe you need to add a nested `raw` property just like the other widgets have it.

Thanks @ferdingler. I will try to break them up better in the future. That is a good point about it being easy to miss stuff.

You are right. Great catch. I'm not sure how I missed that. I guess it was hard to do a full end to end test when just doing small parts like this.

frontend/src/containers/AddImage.tsx
ajmokotoff: Great catch. Fixed

ferdingler (approved):

PR #253, opened 2021-01-26
GTT 298 - Add new screens to add metrics widget

miguelpdiaz8:
## Description

Add new screens to add metrics widget. This includes the DatePicker and MetricsCardGroup components. Edit will be added in a separate PR.

## Testing

Tested it locally here: https://migdizn.badger.wwps.aws.dev/admin/dashboard/4d041da6-0fb3-48dc-8294-5a0d1bc46e45/add-content

---


backend/src/lib/factories/widget-factory.ts
```diff
@@ -243,6 +243,7 @@ function createMetricsWidget(widget: Widget): MetricsWidget {
     content: {
       title: widget.content.title,
       datasetId: widget.content.datasetId,
+      oneMetricPerRow: widget.content.oneMetricPerRow,
```
ferdingler, line 246: May need to update the tests to verify this property?

frontend/src/components/MetricsCardGroup.tsx
```diff
@@ -0,0 +1,90 @@
...
+                                metric.changeOverTime[0] === "-"
+                                  ? faArrowDown
+                                  : faArrowUp
+                              }
+                              className="margin-right-2px"
+                            />
+                            {metric.changeOverTime.substring(1)}
+                          </div>
+                        ) : (
+                          ""
```
ferdingler, line 62: In these cases I prefer to use this syntax:

```
{metrics.startDate && (<somehtml />)}
```

It avoids the need for adding the empty "" in the `else` block.

frontend/src/containers/EditMetric.css
```diff
@@ -0,0 +1,3 @@
+.width-50 {
```
ferdingler, line 1: Looks like this is being defined in multiple places, can we put it in the index.css instead?

frontend/src/services/StorageService.ts
```diff
@@ -113,6 +113,20 @@ async function uploadDataset(
   };
 }
 
+async function uploadMetric(jsonFile: string): Promise<UploadDatasetResult> {
```
ferdingler, line 118: An alternative more generic name for this function could be `uploadJson`, because there is nothing specific about metrics in this function other than simply uploading a JSON file.

ajmokotoff: This function calls a function named uploadJson already.  Maybe we should discuss this module later, but I essentially left out the s3 key generation logic outside of the uploadJson function since it seemed like the behavior could change among different items that were needed to upload. 

For example: dataset - the keys were shared for the uploadJson and uploadFile. So I thought it made sense to make a function called "uploadDataset" that would just call uploadFile + uploadJson with the keys passed in. @miguelpdiaz8 is following the pattern that I started.

Should we all discuss later if we should refactor?

miguelpdiaz8: Yes, as Andrew said we already have a function that specifically upload a json to S3. This new function uploadMetric encapsulate the key generation and then upload the json to S3 using the uploadJson function. We can discuss later.

ferdingler (approved):

PR #260, opened 2021-01-28
BUG Fix: Allow blob images to appear on site

ajmokotoff:
## Description

Allow blob images to appear on the site. The content security policy was blocking this. Now images will only be allowed from our site (which was the default already) but now blobs will be allowed as well. So all images + blob images, will be allowed from our own site.

## Testing

I deployed to my AWS instance and verified that the error was gone and that images were showing up.

---


trietlu (approved):
lgtm

PR #273, opened 2021-02-04
GTT-894: Modified the file input component to meet the UX wireframe

ajmokotoff:
## Description

This ticket was to meet the wireframe for what the file input component looks like when a file is selected. If you go to this [link](https://designsystem.digital.gov/components/form-controls/#file-input), you will see that by default this component matches the wireframe. However, because this component is using javascript to hide and show different classes, this does not work with react. Since with react you cannot modify the dom once its displayed. The dom will only change if you modify a state attribute.

Upon inspecting the site, you can see the classes that it is toggling in order to display the image like the wireframe. In this change, I manually show these classes as a function of the file state variable. I also added come CSS to get the label and file logo to align properly. In addition, in order to get the file icon, I used font awesome.

Only concerns are that we are adding additional dependency on external CSS. This component will already break if the CSS changes, this change would be adding additional dependency on the external CSS. We would need to be careful with updating uswds. 

Below is what the interface looks like now.

Screen Shot 2021-02-03 at 20 49 25

## Testing

Tests pass, initial tests to the front end look good. 

---

### DIFF
--- a/frontend/src/components/FileInput.css
+++ b/frontend/src/components/FileInput.css
@@ -1,0 +1,10 @@
+.fileIcon {
+  display: inline-block;
+  height: 0%;
+  vertical-align: middle;
+  padding-left: 30%;
+}
+
+#fileIconLabel {
+  margin-top: 0;
+}
--- a/frontend/src/components/FileInput.css
+++ b/frontend/src/components/FileInput.css
@@ -4,7 +4,3 @@
   vertical-align: middle;
   padding-left: 30%;
 }
-
-#fileIconLabel {
-  margin-top: 0;
-}
--- a/frontend/src/components/FileInput.tsx
+++ b/frontend/src/components/FileInput.tsx
@@ -1,5 +1,8 @@
 import React from "react";
 import Spinner from "./Spinner";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faFile } from "@fortawesome/free-solid-svg-icons";
+import "./FileInput.css";
 
 interface Props {
   id: string;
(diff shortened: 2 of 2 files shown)


frontend/src/components/FileInput.css
```diff
@@ -0,0 +1,10 @@
+.fileIcon {
+  display: inline-block;
+  height: 0%;
+  vertical-align: middle;
+  padding-left: 30%;
+}
+
+#fileIconLabel {
+  margin-top: 0;
+}
```
ferdingler: No need to create this class. You can use the Utility classes that USWDS already has: https://designsystem.digital.gov/utilities/margin-and-padding. 

In this case, you should be able to add `margin-top-0` to your label: 
```jsx
<label className="usa-label margin-top-0" htmlFor="fileIconDiv">
   {props.fileName}
</label>
```

ferdingler (changes requested):
Also, what do you mean when you say that we are "adding additional dependency on external CSS"? What you did here seems fine to me. I'm not sure I understand why this would break if USWDS changes.

ajmokotoff:
> Also, what do you mean when you say that we are "adding additional dependency on external CSS"? What you did here seems fine to me. I'm not sure I understand why this would break if USWDS changes.

Sorry, maybe I was being confusing. But for example, if usa-file-input__preview changes in the CSS from USWDS, it could potentially break our UI. Since I did some CSS around it.

frontend/src/components/FileInput.css
ajmokotoff: good call.

ferdingler (approved):

PR #274, opened 2021-02-04
Change X,Y axis to number type if all row/column values are numbers

miguelpdiaz8:
## Description

Change X,Y axis to number type if all row/column values are numbers

## Testing

---

### DIFF
--- a/frontend/src/components/BarChartPreview.tsx
+++ b/frontend/src/components/BarChartPreview.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useCallback } from "react";
 import {
   XAxis,
   YAxis,
@@ -15,12 +15,19 @@ type Props = {
   title: string;
   summary: string;
   bars: Array<string>;
-  data?: Array<object>;
+  data?: Array<any>;
   summaryBelow: boolean;
 };
 
 const BarChartPreview = (props: Props) => {
   const colors = useColors(props.bars.length);
+  const yAxisType = useCallback(() => {
+    return props.data &&
+      props.data.every((row) => typeof row[props.bars[0]] === "number")
+      ? "number"
+      : "category";
+  }, [props]);
+
   return (
     <div>
       <h2 className="margin-left-1 margin-bottom-1">{props.title}</h2>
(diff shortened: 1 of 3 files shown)


frontend/src/components/BarChartPreview.tsx
```diff
@@ -15,12 +15,19 @@ type Props = {
...
 
 const BarChartPreview = (props: Props) => {
   const colors = useColors(props.bars.length);
+  const yAxisType = useCallback(() => {
+    return props.data &&
+      props.data.every((row) => typeof row[props.bars[0]] === "number")
+      ? "number"
+      : "category";
+  }, [props]);
+
```
ferdingler: Soon we need to start moving some of these logic to a `DatasetParsing.ts` file or something like that, so that we don't repeat this logic in every Chart component.

ferdingler (dismissed):

frontend/src/components/ColumnChartPreview.tsx
```diff
@@ -14,12 +14,19 @@ type Props = {
...
 };
 
 const ColumnChartPreview = (props: Props) => {
   const colors = useColors(props.columns.length);
+  const xAxisType = useCallback(() => {
+    return props.data &&
+      props.data.every((row) => typeof row[props.columns[0]] === "number")
+      ? "number"
+      : "category";
+  }, [props]);
```
ferdingler: It's more efficient if we set specifically `data` and `columns` as dependencies of the `useCallback` instead of the entire `props` object. We can avoid some unnecessary re-renders by doing something like this: 

```javascript
const { data, columns } = props;
const xAxisType = useCallback(() => {
    return data &&
      data.every((row) => typeof row[columns[0]] === "number")
      ? "number"
      : "category";
  }, [data, columns]);
```

ferdingler (changes requested):

frontend/src/components/ColumnChartPreview.tsx
miguelpdiaz8: Sounds good, done

ferdingler (approved):

PR #275, opened 2021-02-04
Bug fix: Public users can now see images

ajmokotoff:
## Description

I added image file types to the public user policy, so they don't get 403 errors on the front end.

## Testing

I deployed to my AWS account, and published a dashboard, and when into an incognito window, and tried viewing it as a public user. It worked.

---

### DIFF
--- a/cdk/lib/auth-stack.ts
+++ b/cdk/lib/auth-stack.ts
@@ -127,7 +127,12 @@ export class AuthStack extends cdk.Stack {
       new iam.PolicyStatement({
         effect: iam.Effect.ALLOW,
         actions: ["s3:GetObject"],
-        resources: [bucketArn.concat("/public/*.json")],
+        resources: [
+          bucketArn.concat("/public/*.json"),
+          bucketArn.concat("/public/*.png"),
+          bucketArn.concat("/public/*.jpg"),
+          bucketArn.concat("/public/*.svg"),
+        ],
       })
     );
 


cdk/lib/auth-stack.ts
```diff
@@ -127,7 +127,12 @@ export class AuthStack extends cdk.Stack {
       new iam.PolicyStatement({
         effect: iam.Effect.ALLOW,
         actions: ["s3:GetObject"],
-        resources: [bucketArn.concat("/public/*.json")],
+        resources: [
+          bucketArn.concat("/public/*.json"),
+          bucketArn.concat("/public/*.png"),
+          bucketArn.concat("/public/*.jpg"),
```
ferdingler, line 133: do we need to add `jpeg` as well?

ajmokotoff: > do we need to add `jpeg` as well?

Jpegs work. I tested all 4 types. I could add one more line if you want to be more explicit. But I think jpg allows both jpg and jpeg.

ferdingler: I don't think it will work if a file is named `banana.jpeg`. The IAM policy matches the exact name of the object including the extension.

ferdingler (approved):
Verified with @ajmokotoff that JPG images are always saved as `.jpg` on S3. We don't need to add `.jpeg` to the IAM policy.

PR #280, opened 2021-02-06
GTT-927 Remove users api

ajmokotoff:
## Description

Made some initial progress but did not test anything yet. Here is a head start if it will help you guys/

## Testing

---

### DIFF
--- a/backend/src/lib/controllers/user-ctrl.ts
+++ b/backend/src/lib/controllers/user-ctrl.ts
@@ -1,7 +1,10 @@
 import { Request, Response } from "express";
 import UserRepository from "../repositories/user-repo";
 import UserFactory from "../factories/user-factory";
+import { validate } from "jsonschema";
+import RemoveUsersSchema from "../jsonschema/api/RemoveUsers.json";
 import { Role } from "../models/user";
+import logger from "../services/logger";
 
 async function getUsers(req: Request, res: Response) {
   const repo = UserRepository.getInstance();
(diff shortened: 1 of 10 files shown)


backend/src/lib/controllers/user-ctrl.ts
```diff
@@ -51,6 +54,24 @@ async function addUsers(req: Request, res: Response) {
   }
 }
 
+async function removeUsers(req: Request, res: Response) {
+  const validationResult = validate(req.body, RemoveUsersSchema);
+  if (!validationResult.valid) {
+    res.status(400);
+    logger.warn("Invalid request to remove users %o", validationResult.errors);
+    return res.send(validationResult.toString());
+  }
```
ferdingler, line 63: @miguelpdiaz8 Let's start doing validation in the Backend using jsonschema like this. What do you think?

ferdingler: @trietlu

backend/src/lib/jsonschema/api/RemoveUsers.json
```diff
@@ -0,0 +1,14 @@
...
+  "properties": {
+    "usernames": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    }
+  },
+  "required": ["usernames"]
+}
```
ferdingler, line 14: @trietlu @miguelpdiaz8 We would need to define the json schema for the Request object for each of our endpoints like this. The nice thing is that we should be able to leverage these schemas when we build our Open API documentation since it supports `jsonschema`.

backend/src/lib/controllers/user-ctrl.ts
miguelpdiaz8: SGTM

miguelpdiaz8 (approved):

backend/src/lib/controllers/user-ctrl.ts
trietlu: @ferdingler this is great!

PR #288, opened 2021-02-11
Bug: Fix issue when switching Part-To-Whole charts dataset types

miguelpdiaz8:
## Description

Bug: Fix issue when switching Part-To-Whole charts dataset types

## Testing

Tested it in my local environment.

---

### DIFF
--- a/frontend/src/components/PartWholeChartPreview.tsx
+++ b/frontend/src/components/PartWholeChartPreview.tsx
@@ -29,7 +29,7 @@ const PartWholeChartPreview = (props: Props) => {
 
   const { data, parts } = props;
   useMemo(() => {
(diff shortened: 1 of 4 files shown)


trietlu (commented):
@miguelpdiaz8 What's the purpose of the timeout?

miguelpdiaz8:
> @miguelpdiaz8 What's the purpose of the timeout?

This is to wait until the next tick. That allow react to re-render the PartToWhole chart, what is happening now is the chart being rendered using the previous chart layout instead of the new one. Waiting a tick fix the issue and we are not waiting even a second, so the performance wont be affected.

trietlu (approved):
lgtm

PR #290, opened 2021-02-11
GTT-926 Implement SSO with SAML Identity Provider

trietlu:
## Description

**This PR is to enable a code review, and will not be merged into mainline**.  After reviewing, the changes will remain in a feature branch until the feature has been approved by Security.  After that, it will be merged into mainline

GTT-926, GTT-871

* Replaced withAuthenticator with withSAMLAuthenticator to expose "enterprise sign-in" button which calls federatedSignIn() to Identity Provider (IdP)
*  Added properties to EnvConfig related to federated login
*  Added properties to amplify-config required by Amplify federatedSignIn
*  Updated user-hooks to look for ["custom:groups"] in addition to ["custom:roles"] in Id token when doing federated authentication
* Updated auth.ts to look for ["custom:groups"] in Id token in addition to ["custom:roles"]
*  Modified logout to do single sign-out

## Testing

Tested in my AWS [environment](https://trietlu.badger.wwps.aws.dev/admin)
* login and logout as federated users with Admin and Editor roles (separately)
* login and logout as Cognito user
* tested single sign-out

Will be adding automated API test in another PR

---


backend/src/lib/repositories/user-repo.ts
```diff
@@ -34,9 +34,12 @@ class UserRepository {
...
 
-    return result.Users.map((cognitoUser) =>
-      UserFactory.fromCognitoUser(cognitoUser)
-    );
+    return result.Users.reduce(function (result, cognitoUser) {
+      if (cognitoUser.UserStatus !== "EXTERNAL_PROVIDER") {
+        result.push(UserFactory.fromCognitoUser(cognitoUser));
+      }
+      return result;
+    }, [] as User[]);
```
ferdingler: Nit pick: Chaining `filter` + `map` might be more readable than `reduce`. Reduce is often used to return a single number (like a Sum of the elements in an array). In this case, you are trying to filter out elements that don't match a given condition, for which `filter` seems like a more natural fit. 

```js
return result.Users
  .filter(cognitoUser => cognitoUser.UserStatus !== "EXTERNAL_PROVIDER")
  .map(UserFactory.fromCognitoUser);
```

ferdingler (commented):
A couple of questions: 

1) Does the "Enterprise Sign-In" button always shows up even if the customer hasn't configured federation? 

2) At what point are we deleting the user "clone" to make sure that users that are disabled in the IdP no longer exist in the Cognito User Pool?

backend/src/lib/repositories/user-repo.ts
trietlu: @ferdingler Good suggestion.  I was thinking filter and map will loop through twice versus once for reduce

cdk/lib/auth/pretokengeneration/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    claimsOverrideDetails: {
+      claimsToAddOrOverride: {
+        "custom:roles": roles,
+      },
+    },
+  };
+  console.log(event.response);
+
+  return event;
+};
```
ferdingler: I'm worried about the complexity that we are adding to our Auth workflow here. I know we have discussed this before but we should consider offloading this complexity to Cognito. What I mean is that we shouldn't create a record for a federated user in our Cognito User Pool. Let Cognito manage those records for external identities. Having to maintain this "user clone" is opening the door for issues.

ferdingler: The more agnostic our application can be to Federation, the better. Let customers configure federation in Cognito if they want, but our application shouldn't need to change or know if the users are coming from an external identity provider or from Cognito.

backend/src/lib/repositories/user-repo.ts
trietlu: @ferdingler In EnvConfig, there's now a SAML Provider field, which is the name of the IdP we configured in Cognito.  If that field is empty, then everything is as it was before the federation, which means there's no "Enterprise Sign-In" button

We don't delete the cloned Cognito user because it will be created again the next time the user logs in.  That cloned user doesn't have a role, and can't be used to login to Badger.

cdk/lib/auth/pretokengeneration/index.ts
trietlu: @ferdingler That's what I recommended initially, that we don't manage users in Badger at all.  But since we create users and roles in Badger, then we have to map the federated user to the Badger user.  I believe Quicksights does the same

trietlu: That's correct, customers can configure federation in Cognito, and it wouldn't affect Badger.  But once they set EnvConfig to point Badger to use federation, then Badger must know to do a federated sign in, which is a different flow

PR #296, opened 2021-02-16
GTT-987: Create S3 Bucket and Configure Permissions for Custom Content (Add Logo)

ajmokotoff:
## Description

Create S3 bucket and configure permissions for custom contnt (add logo).

## Testing

Deployed locally, and went onto my own AWS instance, and verified the bucket was created.

---


cdk/bin/main.ts
```diff
@@ -19,6 +19,7 @@ const stackPrefix = "PerformanceDash-".concat(envName);
 const accountId = cdk.Aws.ACCOUNT_ID;
 const region = cdk.Aws.REGION;
 const datasetsBucketName = `performancedash-${envName.toLowerCase()}-${accountId}-${region}-datasets`;
+const contentBucketName = `performancedash-${envName.toLowerCase()}-${accountId}-${region}-customcontent`;
```
ferdingler: Bucket names can be up to 63 characters long. Should we name the bucket just `content` instead of `customcontent` to shorten it by 6 characters and leave more room to the Environment name? I'm afraid that if a customer enters a long Environment Name, the bucket will exceed the 63 chars and fail the deployment. 

What do you think?

ajmokotoff: Good call. I also name it content everywhere else. So I think that makes sense.

ferdingler (approved):

PR #298, opened 2021-02-17
GTT-988: Backend for Logo Upload

ajmokotoff:
## Description

Added backend support to save the S3 ID for custom logos in the settings table.

## Testing

Wrote additional tests and verified that all tests passed.

---


backend/src/lib/models/settings.ts
```diff
@@ -10,6 +10,7 @@ export interface Settings {
     singular: string;
     plural: string;
   };
+  customLogoS3ID: string | undefined;
```
ajmokotoff: I wasn't sure whether to make this field optional, instead of forcing it to exist and be either a string or undefined. My thoughts are that it would make the code messier if we made the field optional. In this case, we will make it undefined by default, and once the user uploads a logo, it will be populated from that point on.

Unless there was an easy way to make the default logo we have now, populated in the S3 bucket upon creation of the badger environment, and by default, this field points to that S3 ID. Thoughts? @ferdingler @miguelpdiaz8

backend/src/lib/models/settings.ts
```diff
@@ -22,6 +23,7 @@ export interface PublicSettings {
...
```
ajmokotoff, line 26: This might not need to be here.

ajmokotoff: I'm going to remove it.

backend/src/lib/models/settings.ts
ferdingler: I agree that this field should be optional. If undefined, we'll show the default logo. I'm not aware of an easy way to populate the bucket with the current logo at deployment time, so your approach sounds easier.

ferdingler: I would however call this property `customLogoS3Key`. The proper term is `Key` to refer to an S3 item and we have been using `S3Key` for the datasets as well so I suggest we stick to that naming.

ajmokotoff: I was thinking on the front end, if that field was undefined, just grabbing the image from the current path, instead of S3.

ferdingler (approved):

PR #300, opened 2021-02-18
Fix issue with x, y axes being cut off with long labels

miguelpdiaz8:
## Description

Fix issue with x, y axes being cut off with long labels

## Testing

Tested it in my local environment. You can test it here: https://migdizn.badger.wwps.aws.dev/admin/dashboard/ab13ccfa-17e5-4f5f-826d-0c628a621edc/edit-chart/f3b20700-fa48-4c54-8004-43b675dbd071

---

### DIFF
--- a/frontend/src/components/BarChartPreview.tsx
+++ b/frontend/src/components/BarChartPreview.tsx
@@ -11,6 +11,7 @@ import {
   Tooltip,
 } from "recharts";
 import { useColors } from "../hooks";
+import UtilsService from "../services/UtilsService";
 
 type Props = {
   title: string;
@@ -51,6 +52,9 @@ const BarChartPreview = (props: Props) => {
     }
   };
 
+  const formatYAxisLabel = (label: string) =>
+    label.length > 27 ? label.substr(0, 27).concat("...") : label;
+
   return (
     <div>
       <h2
@@ -65,10 +69,10 @@ const BarChartPreview = (props: Props) => {
           {props.summary}
         </p>
       )}
-      {props.data && props.data.length && (
+      {data && data.length && (
         <ResponsiveContainer
           width="100%"
-          height={props.data && props.data.length > 15 ? 600 : 300}
+          height={data && data.length > 15 ? 600 : 300}
         >
           <BarChart
             data={props.data}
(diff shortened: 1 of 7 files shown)


frontend/src/components/BarChartPreview.tsx
```diff
@@ -86,18 +90,13 @@ const BarChartPreview = (props: Props) => {
...
-                  )
-                  .map((c) => (c as string).length)
-                  .reduce((a, b) => (a > b ? a : b), 0) || 0) *
-                  8 +
-                24
-              }
+              width={Math.min(
+                UtilsService.getLargestHeader(props.bars, props.data) * 8 + 24,
+                220
+              )}
```
ferdingler: It feels like lately we have been adding a lot of hardcoded "magical" numbers on the Chart components. Is there any way we can make this better?

frontend/src/components/ColumnChartPreview.tsx
```diff
@@ -50,8 +52,15 @@ const ColumnChartPreview = (props: Props) => {
     }
   };
 
+  const widthPercent =
+    (UtilsService.getLargestHeader(columns, data) *
+      (data ? data.length : 0) *
+      8 *
+      100) /
+    (props.isPreview ? 480 : 960);
```
ferdingler: Similar comment. `480`, `960`, `8 * 100`, `0`. What do these numbers mean? Are we going to remember in the future? What about making them constants on the top of the file to give them descriptive names at least? Would that help?

frontend/src/components/BarChartPreview.tsx
miguelpdiaz8: Yeah, I agree, those are kind of magic numbers, I had to spend a lot of time figuring that out, and I dont think it is perfect yet. The problem is recharts does not support some functionalities, so we need to do the calculations manually. One way I see we could do it better is to create variables for each number describing what they represent (offset, pixel by character, etc)

frontend/src/components/ColumnChartPreview.tsx
miguelpdiaz8: Same

ferdingler (approved):

PR #307, opened 2021-02-23
GTT 989: Edit Logo Feature

ajmokotoff:
## Description

This is the front end portion of the new edit logo feature. 

## Testing

I wrote tests and tested locally.

---


frontend/src/containers/BrandingAndStylingSettings.tsx
```diff
@@ -0,0 +1,48 @@
...
+      </p>
+
+      <br></br>
+
+      <div className="grid-row margin-top-0-important">
+        <div className="grid-col flex-1 text-left">
+          <Logo />
+        </div>
+        <div className="grid-col flex-9">
+          &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
```
ajmokotoff: There is probably a better way to do this.

miguelpdiaz8 (commented):
Can you share your environment to see how it looks? Thanks

frontend/src/components/Logo.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { settings, loadingSettings } = useSettings();
+  const { logo, loadingFile } = useLogo(settings.customLogoS3Key);
+
+  return (
+    <>
+      {loadingFile || loadingSettings ? (
+        <Spinner className="margin-top-3 text-center" label="Loading" />
+      ) : (
+        <>
+          <img src={logo ? URL.createObjectURL(logo) : defaultLogo}></img>
```
ferdingler: Let's add an `alt` attribute for accessibility: 

```html
<img src="..." alt="Organization logo" />
```

frontend/src/containers/BrandingAndStylingSettings.tsx
miguelpdiaz8: Yes, you could add a left margin to the button with the USWDS class margin-left-1 or another value.

ferdingler: yeah, just add padding left class from USWDS to your div. For example: 
```html
<div className="padding-left-3">
```

frontend/src/containers/EditLogo.tsx
```diff
@@ -0,0 +1,123 @@
...
+        setImageUploading(false);
+
+        history.push("/admin/settings/brandingandstyling", {
+          alert: {
+            type: "success",
+            message: "Logo successfully modified",
+          },
+        });
+      } catch (err) {
+        console.log("Failed to save logo", err);
```
ferdingler: Show an error alert if this happens instead of console log?
```js
history.push("/admin/settings/brandingandstyling", {
          alert: {
            type: "error",
            message: "Failed to update logo",
          },
});
```

frontend/src/hooks/logo-hooks.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState, useCallback } from "react";
+import ContentService from "../services/ContentService";
+
+type UseLogoHook = {
+  loadingFile: boolean;
+  logo: File | undefined;
+};
+
+export function useLogo(s3Key: string | undefined): UseLogoHook {
```
ferdingler: I think is the same if you make the parameter optional `?`. Removes the need to specify `undefined` as an accepted data type. just fyi, no need to change. 
```js
function useLogo(s3Key?: string) {
  
}
```

frontend/src/layouts/Admin.css
```diff
@@ -0,0 +1,4 @@
+logo {
+  height: 2rem;
+  width: 2rem;
+}
```
ferdingler: This seems to restrict the logo to be a square? Have we tried with other logos that are rectangular?

frontend/src/models/index.tsx
```diff
@@ -195,6 +195,7 @@ export type PublicSettings = {
     singular: string;
     plural: string;
   };
+  customLogoS3Key: string | undefined;
```
ferdingler: similar comment, adding the `?` makes the property optional which is the same as saying it may be `undefined`.

frontend/src/services/ContentService.ts
```diff
@@ -0,0 +1,82 @@
...
+  return await uploadImage(rawFile, "logo");
+}
+
+const StorageService = {
+  downloadLogo,
+  uploadLogo,
+  imageFileTypes,
+};
+
+export default StorageService;
```
ferdingler: Was there no way to reuse the existing StorageService we already had? Perhaps just adding additional functions that are specific to uploading to the content bucket? This way we don't have 2 services that look exactly the same.

ferdingler (dismissed):
Minor comments

frontend/src/containers/EditLogo.tsx
ajmokotoff: Good call.

frontend/src/hooks/logo-hooks.ts
ajmokotoff: ah nice. That is cleaner. But mine is more verbose. I guess we'll stick with mine?

frontend/src/layouts/Admin.css
ajmokotoff: This appeared to be the constraint on the old placeholder logo. I have used rectangular logos, and it doesn't change the size of the header.

frontend/src/services/ContentService.ts
ajmokotoff: We can do that. I thought it might be cleaner to have separate services that interacted with their own respective S3 bucket. But if you think it would be cleaner to merge the two, I can do that.

miguelpdiaz8 (approved):

PR #308, opened 2021-02-23
GTT-875: Frontend for color settings

miguelpdiaz8:
## Description

Create edit colors page.

## Testing

Added unit tests. You can test it here: https://migdizn.badger.wwps.aws.dev/admin/settings/topicarea/editcolors

---


frontend/src/App.tsx
```diff
@@ -114,6 +115,10 @@ const routes: Array<AppRoute> = [
     path: "/admin/settings/topicarea/editlabel",
     component: EditTopicAreaLabel,
   },
+  {
+    path: "/admin/settings/topicarea/editcolors",
```
ferdingler: Looks like a typo, did you mean to put the colors inside a `topicarea`?

frontend/src/components/LineChartPreview.tsx
```diff
@@ -19,12 +19,29 @@ type Props = {
...
+  useEffect(() => {
+    if (settings && settings.colors) {
+      const colors = UtilsService.getColors(
+        props.lines.length,
+        props.colors?.primary || settings.colors.primary,
+        props.colors?.secondary || settings.colors.secondary
+      );
+      setChartColors(colors);
+    }
+  }, [settings, props.lines, props.colors]);
```
ferdingler: potential opportunity to abstract into a custom hook. I see these lines repeated in the other Widgets. A separate PR later for the refactor is fine.

frontend/src/components/__tests__/ColumnChartPreview.test.tsx
```diff
@@ -3,6 +3,8 @@ import { render } from "@testing-library/react";
 import { MemoryRouter } from "react-router-dom";
 import ColumnChartPreview from "../ColumnChartPreview";
 
+jest.mock("../../hooks");
```
ferdingler: Hooks are actually automaticallly mocked because the `setupTests.ts` file already does it. This line doesn't affect but it shouldn't be necessary.

frontend/src/hooks/index.tsx
```diff
@@ -29,7 +29,6 @@ export {
   usePublicDashboard,
   useDashboards,
   useWidget,
-  useColors,
```
ferdingler: I wonder if it was necessary to get rid of the `userColors` hook? I understand that the colors are now coming from the Settings object but you could have used the `useSettings` inside the `useColors` hook to avoid such a big refactor.

frontend/src/services/UtilsService.ts
```diff
@@ -60,12 +64,54 @@ function getLargestHeader(headers: Array<string>, data?: Array<any>) {
...
+  }
+  return colors;
+};
+
+const getSecondaryOptions = () => {
+  return defaultColors.map((c) => {
+    return { value: c, content: c };
+  });
+};
+
```
ferdingler: We need to be cautious about our UtilsService growing too much. It's very tempting to just throw functions into Utils but it shouldn't be our default place to put them. In this case, the colors were already part of a `useColors` hook which was a good place to be. I don't think the UtilsService is the right place for them.

ferdingler: In my opinion, Utils should be very specific functions that perform "plumbing" work that is not really part of our Domain or business logic. Example: Functions to manipulate an array or a string. In this case, the Color Palette is something important in our core domain and I think it deserves it's own service `ColorPaletteService` or to put it back into a `useColors` hook. 

What do you think?

ferdingler (changes requested):

frontend/src/App.tsx
miguelpdiaz8: This is because I'm waiting on Andrew PR to merge it, that Settings page is implemented there.

frontend/src/components/LineChartPreview.tsx
miguelpdiaz8: Good idea

frontend/src/services/UtilsService.ts
miguelpdiaz8: Agree, I'll commit the updates shortly, I did not like it the useColors hooks the way it existed but with the Color palette service now it makes sense.

ferdingler (approved):

PR #310, opened 2021-02-24
GTT-551: Modify Edit Details screen to show a live preview

miguelpdiaz8:
## Description

Modify Edit Details screen to show a live preview and add snapshot tests.

## Testing

Added snapshot tests. You can test it here: https://migdizn.badger.wwps.aws.dev/admin/dashboard/edit/80d09d4e-4ec6-4eaa-b0cd-520927816bda/details 

---

### DIFF
--- a/frontend/src/containers/__tests__/TopicareaSettings.test.tsx
+++ b/frontend/src/containers/__tests__/TopicareaSettings.test.tsx
@@ -184,3 +184,10 @@ test("filters topic areas based on createdBy search input", async () => {
   // Topic area one should appear
   expect(topicarea).toBeInTheDocument();
 });
+
+test("topic area settings should match snapshot", async () => {
+  const wrapper = render(<TopicareaSettings />, {
+    wrapper: MemoryRouter,
+  });
+  expect(wrapper.container).toMatchSnapshot();
+});
--- a/frontend/src/containers/BrandingAndStylingSettings.tsx
+++ b/frontend/src/containers/BrandingAndStylingSettings.tsx
@@ -51,6 +51,15 @@ function BrandingAndStylingSettings() {
         </div>
       </div>
 
+      <hr
+        style={{
+          border: "none",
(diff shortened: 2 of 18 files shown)


frontend/src/containers/__tests__/TopicareaSettings.test.tsx
```diff
@@ -184,3 +184,10 @@ test("filters topic areas based on createdBy search input", async () => {
   // Topic area one should appear
   expect(topicarea).toBeInTheDocument();
 });
+
+test("topic area settings should match snapshot", async () => {
+  const wrapper = render(<TopicareaSettings />, {
+    wrapper: MemoryRouter,
+  });
+  expect(wrapper.container).toMatchSnapshot();
+});
```
ferdingler, line 193: I'm not a fan of snapshot testing in Containers. It makes the snapshot fragile to break if any of the underlying components change, which can be many. I think we should try to keep snapshots only for components and use them sporadically. When possible, we should favor basic assertions with react-testing-library. 

What do you think?

miguelpdiaz8: Yes, I was surprised when I noted we dont have snapshot tests for containers. I really dont see a good reason why not to have it. If an underlying component changes I think it is fine to update the containers too, anyway update a snapshot test is fairly easy and the alternative of not having them, avoid to snapshot test almost the whole container. Of course ideally the containers should have only other components but in practice that is not always the case. Another area to improve is to create more components out of the containers, though.

ferdingler (approved):

frontend/src/containers/__tests__/TopicareaSettings.test.tsx
ferdingler: Let's have a team discussion on this. I don't think we need snapshot testing everywhere but let's discuss it.

PR #313, opened 2021-02-25
Translated archived tab

ajmokotoff:
## Description

Translated the archived tab page

## Testing

Verified it worked in my local environment

---

### DIFF
--- a/frontend/src/locales/es/translation.json
+++ b/frontend/src/locales/es/translation.json
@@ -2,6 +2,5 @@
   "LastUpdatedLabel": "Última actualización",
   "SearchButton": "Buscar",
   "ClearSearchText": "Limpiar parámetros de búsqueda",
-  "AllDashboardsLink": "Todos los dashboards",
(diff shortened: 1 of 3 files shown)


frontend/src/locales/es/translation.json
```diff
@@ -2,5 +2,6 @@
   "LastUpdatedLabel": "Última actualización",
   "SearchButton": "Buscar",
   "ClearSearchText": "Limpiar parámetros de búsqueda",
-  "AllDashboardsLink": "Todos los dashboards"
+  "AllDashboardsLink": "Todos los dashboards",
+  "ArchivedTabDescription": "to be translated"
```
ferdingler: Since we don't have the translation yet, a better approach would be to not include this key at all in the Spanish json file. That way it will default to English instead of showing a `to be translated` string.

ferdingler (changes requested):

ferdingler (approved):
@ajmokotoff make sure to update the Quip document to mark this component as translated.

PR #318, opened 2021-02-25
GTT-980: Special characters not encoded properly when in uploaded files

ajmokotoff:
## Description

When you upload a CSV file with special characters such as ü, they appear as unknown characters on badger. This is because papaparse was not configured to handle special characters, as no encoding was set. If you set the encoding to "ISO-8859-1", this allows for csv files made in excel (on a PC) to upload files with speical characters with no problem.

However, there is currently no way to get a csv file made in excel (on a mac) with special characters, to work properly. The only work around for a mac, is for the user in excel to save the file as a UTF-8 CSV file.

This fix allows PCs to upload files with special characters with no issue.

## Testing

I tested with @jasongudalis by deploying to my environment, and having him upload csv files made on his PC with special characters. He was able to successfully upload files with the below fix, and have special characters show up properly.

---

### DIFF
--- a/frontend/src/containers/AddTable.tsx
+++ b/frontend/src/containers/AddTable.tsx
@@ -165,6 +165,7 @@ function AddTable() {
       dynamicTyping: true,
       skipEmptyLines: true,
       comments: "#",
+      encoding: "ISO-8859-1",
       complete: function (results: ParseResult<object>) {
         if (results.errors.length) {
           setCsvErrors(results.errors);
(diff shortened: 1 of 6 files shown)


frontend/src/containers/AddTable.tsx
```diff
@@ -165,6 +165,7 @@ function AddTable() {
       dynamicTyping: true,
       skipEmptyLines: true,
       comments: "#",
+      encoding: "ISO-8859-1",
```
ferdingler, line 168: The fact that you had to make the change in 4 different files speaks about our opportunity to DRY-out the parsing. If you have a chance later this sprint, could you please do the refactor of extracting this parsing function to it's own `CsvParserService`?

ferdingler (approved):

PR #321, opened 2021-02-26
GTT-1032 Publish pending endpoint now allows update of release notes

ferdingler:
## Description

The endpoint to set a dashboard to publish pending now allows to pass `releaseNotes` as an optional property in the request body. This will allow the frontend to save the release notes as many times as necessary without moving the dashboard to published state. 

I also updated the endpoint to return the newly updated dashboard. We should refactor all the other endpoints to do the same. Whenever we update a dashboard via a PUT request, the endpoint should return the updated entry. This will save a GET call to fetch it right after an update. 

## Testing

Ran all the integration tests locally to make sure I didn't break backwards compatibility with the API. I updated the unit tests and also added new ones. I also updated our Postman collection to add the new optional field `releaseNotes` in the request body. 

---

### DIFF
--- a/backend/src/lib/controllers/dashboard-ctrl.ts
+++ b/backend/src/lib/controllers/dashboard-ctrl.ts
@@ -238,7 +238,7 @@ async function publishPendingDashboard(req: Request, res: Response) {
     dashboard.state !== DashboardState.PublishPending
   ) {
     res.status(409);
-    return res.send("Dashboard must be in draft state");
+    return res.send("Dashboard must be in draft or publish pending state");
   }
 
   const updatedDashboard = await repo.publishPendingDashboard(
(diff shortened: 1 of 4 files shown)


backend/src/lib/controllers/dashboard-ctrl.ts
```diff
@@ -224,23 +224,31 @@ async function publishDashboard(req: Request, res: Response) {
...
   const repo = DashboardRepository.getInstance();
-
   const dashboard = await repo.getDashboardById(id);
-  if (dashboard.state !== DashboardState.Draft) {
+  if (
+    dashboard.state !== DashboardState.Draft &&
+    dashboard.state !== DashboardState.PublishPending
+  ) {
     res.status(409);
     return res.send("Dashboard must be in draft state");
```
miguelpdiaz8: Can you update this message due to now we also accept PublishPending dashboard state?

ferdingler: Good catch. Will do

miguelpdiaz8 (approved):

PR #326, opened 2021-02-26
GTT-390 Added dropdown menu to user management screen

jenboydclark:
## Description

Implemented the dropdown menu on the user management page for the delete users and re-send invites features.

## Testing

Wrote tests. You can also see it in action here --> https://dvinr4n9hngnv.cloudfront.net/admin/users

---

### DIFF
--- a/frontend/src/containers/UserListing.tsx
+++ b/frontend/src/containers/UserListing.tsx
@@ -9,7 +9,8 @@ import BackendService from "../services/BackendService";
 import { useHistory } from "react-router-dom";
 import Modal from "../components/Modal";
 import AlertContainer from "./AlertContainer";
-import Tooltip from "../components/Tooltip";
+import { MenuItem } from "@reach/menu-button";
+import DropdownMenu from "../components/DropdownMenu";
 
 function UserListing() {
   const history = useHistory<LocationState>();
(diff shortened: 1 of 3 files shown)


frontend/src/containers/UserListing.tsx
```diff
@@ -9,7 +9,8 @@ import BackendService from "../services/BackendService";
 import { useHistory } from "react-router-dom";
 import Modal from "../components/Modal";
 import AlertContainer from "./AlertContainer";
-import Tooltip from "../components/Tooltip";
+import { MenuItem } from "@reach/menu-button";
+import DropdownMenu from "../components/DropdownMenu";
```
ferdingler: @jenboydclark suggestion: We should encapsulate reach/menu-button into the `<DropdownMenu />` component itself. The fact that we need to import both DropdownMenu and MenuItem everytime we need to use this component is going to make our application tightly coupled with the reach/menu-button library, which we should avoid. 

An easy way to do this would be to create a subcomponent attached to the <DropDownMenu> component. So when we use it it would look something like this: 

```js

const MenuItem = DropdownMenu.MenuItem;

<DropdownMenu>
  <MenuItem>Action 1</MenuItem>
</DropdownMenu>
```

ferdingler (approved):

PR #328, opened 2021-02-27
Refactored organization of widget components

ferdingler:
## Draft Pull Request

Currently, we have 2 components for every widget type, for example: `<TablePreview>` and `<TableWidget>`, `<ImagePreview>` and `<ImageWidget>`, `<ChartPreview>` and `<ChartWidget>`, etc. With this PR, I am proposing a refactor in which we only need 1 component per widget type. 

For demonstration, I have deleted the `<TablePreview>` component in favor of just keeping `<TableWidget>`. For this to work, I have made the WidgetRender component responsible for fetching the JSON dataset from S3. This way, the TableWidget component is agnostic of where the data is coming from (S3 or a local CSV that just got parsed in the browser). 

If you agree with the approach, I can continue refactoring the other components to remove the need from having 2 components per widget type. 

## Testing

Tested locally and verified that the Table widgets are showing up correctly both while rendering a dashboard as well as when creating a Table content item and seeing the preview. 

---


miguelpdiaz8:
It looks good to me.

miguelpdiaz8:
@ferdingler It looks very good, can you fix the conflict files?

ferdingler:
> @ferdingler It looks very good, can you fix the conflict files?

Done.

miguelpdiaz8 (approved):

PR #329, opened 2021-03-01
Added string locale formatting to the charts + table

ajmokotoff:
## Description

I incorporated localized notation in numbers for readability with charts/tables.

## Testing

I created a csv file with very large numbers and tested the functionality of all the graphs/table views, and verified the numbers were formatted.

---

### DIFF
--- a/frontend/src/components/TablePreview.tsx
+++ b/frontend/src/components/TablePreview.tsx
@@ -37,7 +37,7 @@ const TablePreview = (props: Props) => {
                 minWidth: 150,
                 Cell: (props: any) => {
                   const row = props.row.original;
-                  return row[header].toLocaleString() || null;
+                  return row[header] ? row[header].toLocaleString() : null;
                 },
               };
             }),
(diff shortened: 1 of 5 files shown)


ferdingler:
Have you tested a scenario in which a datapoint (tick) is missing from the CSV file. Example: 

```csv
month,revenue
February,5000
March,
April,1000
``` 

Just to make sure it doesn't trigger a nullpointer exception.

frontend/src/components/TablePreview.tsx
```diff
@@ -37,7 +37,7 @@ const TablePreview = (props: Props) => {
                 minWidth: 150,
                 Cell: (props: any) => {
                   const row = props.row.original;
-                  return row[header] || null;
+                  return row[header].toLocaleString() || null;
```
trietlu: Should we check for null for row[header] before calling toLocaleString()

ajmokotoff: I didn't run into any issues when testing, but it can't hurt.

ferdingler (approved):

PR #347, opened 2021-03-10
GTT: 1071 - Added static datasets selection page

ajmokotoff:
## Description

Added static datasets selection page to add charts

## Testing

Wrote unit tests and verified it. See it live here: https://drjun54o2bnsn.cloudfront.net/

---


frontend/src/components/Table.tsx
```diff
@@ -282,4 +302,27 @@ const IndeterminateCheckbox = React.forwardRef<
   return <input type="checkbox" title={title} ref={resolvedRef} {...rest} />;
 });
 
+const IndeterminateRadio = React.forwardRef<
+  HTMLInputElement,
+  {
+    indeterminate?: boolean;
+    title?: string;
+    checked?: boolean;
+    onClick?: any;
```
ferdingler: make the `onClick` type `Function`?

frontend/src/containers/ChooseStaticDataset.tsx
```diff
@@ -0,0 +1,145 @@
...
+    history.goBack();
+  };
+
+  const onSelect = useCallback((selectedDataset: Array<Dataset>) => {
+    setSelected(selectedDataset);
+  }, []);
+
+  const onSubmit = () => {
+    history.replace({
+      pathname: `/admin/dashboard/${dashboardId}/add-chart`,
```
ferdingler: How are we going to reuse this screen for other types of content items? I think this screen needs to be agnostic of what type of content item the user is adding. Should be easy to do though, just have this screen receive the `redirectUrl` as part of the state, and then when the user selects a dataset, you redirect them back to where they came from.

frontend/src/containers/__tests__/ChooseStaticDataset.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  });
+  expect(
+    screen.getByRole("button", { name: "Select and continue" })
+  ).toBeInTheDocument();
+});
+
+test("renders a cancel button", () => {
+  render(<ChooseStaticDataset />, {
+    wrapper: MemoryRouter,
+  });
```
ferdingler: Nit pick: This render instruction is repeated on every test. You can extract it into a beforeEach: 

```js
beforeEach(() => {
  render(<ChooseStaticDataset />, {
    wrapper: MemoryRouter,
  });
});
```

frontend/src/containers/__tests__/ChooseStaticDataset.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+});
+
+test("renders search box", () => {
+  render(<ChooseStaticDataset />, {
+    wrapper: MemoryRouter,
+  });
+  expect(screen.getByRole("searchbox", { name: "Search" })).toBeInTheDocument();
+});
+
+test("renders a selet and continue button", () => {
```
ferdingler: typo in `select`

frontend/src/containers/__tests__/ChooseStaticDataset.test.tsx
ferdingler: Adding another test to verify that a Table is rendered with the dummy datasets from your mocked hook would be good too.

ferdingler (changes requested):

frontend/src/containers/ChooseStaticDataset.tsx
ajmokotoff: That is a very good point. Let me fix that.

ferdingler (approved):

PR #351, opened 2021-03-12
GTT-1067: Added support for the dynamic datasets

ajmokotoff:
## Description

Added support for dynamic datasets. Enhanced the API to take in additional fields (tags and description).

## Testing

See my local environment here: https://drjun54o2bnsn.cloudfront.net/

My only issue testing was that I was unable to get Postman working. So we will need to test in Gamma to make sure the API works perfectly, 

---

### DIFF
--- a/backend/src/lib/controllers/ingestapi-ctrl.ts
+++ b/backend/src/lib/controllers/ingestapi-ctrl.ts
@@ -55,7 +55,6 @@ async function createDataset(req: Request, res: Response) {
     await repo.saveDataset(dataset);
     res.json(dataset);
   } catch (err) {
-    console.log(err);
     logger.error("Failed to create dataset %o, %o", metadata, parsedData);
     res.status(400).send("Unable to create dataset");
   }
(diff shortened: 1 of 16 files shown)


ferdingler (dismissed):
I don't think we should be adding Tags on this Sprint. We haven't discussed this feature at all. This may require a change in our data model to support searching datasets based on tags.

miguelpdiaz8:
> I don't think we should be adding Tags on this Sprint. We haven't discussed this feature at all. This may require a change in our data model to support searching datasets based on tags.

@ajmokotoff can you remove the backend changes and commit only the support for dynamic datasets in the frontend?

backend/src/lib/controllers/ingestapi-ctrl.ts
```diff
@@ -55,6 +55,7 @@ async function createDataset(req: Request, res: Response) {
     await repo.saveDataset(dataset);
     res.json(dataset);
   } catch (err) {
+    console.log(err);
```
miguelpdiaz8: Can you remove this?

miguelpdiaz8 (approved):

PR #361, opened 2021-03-15
Bug fixes for add chart

ajmokotoff:
## Description

Fixed disappearing charts and made uploading charts with static datasets work.

## Testing

Pair tested with Miguel.

---

### DIFF
--- a/frontend/src/containers/AddChart.tsx
+++ b/frontend/src/containers/AddChart.tsx
@@ -51,7 +51,7 @@ function AddChart() {
   const { dashboardId } = useParams<PathParams>();
   const dateFormatter = useDateTimeFormatter();
   const { dashboard, loading } = useDashboard(dashboardId);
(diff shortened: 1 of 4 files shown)


frontend/src/containers/AddChart.tsx
```diff
@@ -320,12 +307,13 @@ function AddChart() {
...
-      setStep(1);
-      onSelectStaticDataset();
+  useEffect(() => {
+    if (datasetType) {
+      reset({
+        datasetType,
+      });
     }
-  }
+  }, []);
```
ferdingler, line 316: Shouldn't the `datasetType` variable be inside the dependency array?

miguelpdiaz8: We only want to call reset after the initial render

miguelpdiaz8 (approved):

PR #371, opened 2021-03-17
fix: User list wasn't refreshed after deleting users

ferdingler:
## Description

After deleting a user, the success alert will show up but the UI will still show the deleted user. I was able to reproduce this issue by calling the backend directly to delete a user and immediately request the list of users again. I confirmed that the deleted user is still part of the users list. Which leads me to believe that Cognito is eventually consistent, the users are actually deleted but reading them immediately after deleting doesn't show accurate data. 

My solution was to optimistically update the UI to remove the deleted users only if we succeed to delete them from Cognito. If we fail to delete them, then we simply reload the list as we did before. 

## Testing

Tested locally to make sure the users list is updated correctly after deletion. 

---

### DIFF
--- a/frontend/src/containers/UserListing.tsx
+++ b/frontend/src/containers/UserListing.tsx
@@ -43,7 +43,14 @@ function UserListing() {
             message: `Successfully removed ${selected.length} users`,
           },
         });
+
+        // Reload users but also update the UI optimistically because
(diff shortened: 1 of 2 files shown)


trietlu (approved):
Nice approach with the optimistic delete

PR #372, opened 2021-03-18
Add end-to-end tests using cypress

ferdingler:
## Description

Wrote a basic e2e test suite with Cypress. The workflows being tested so far are: 

- Admin user can add a Text content item to a dashboard
- Admin user can add a Metrics content item to a dashboard
- Admin user can add a Line Chart content item to a dashboard
- Admin user can create a new dashboard and delete it
- Admin user can create a user and delete it
- Public users can visit the homepage and see dashboards grouped by topic area

## Testing

I ran the tests several times against gamma and they seem to be pretty reliable. Let's continue to experiment with them. 

---


ajmokotoff (commented):
In file: e2e-tests/cypress/integration/users.spec.ts

Github is flagging random.word() as insecure randomness. I guess this is because you can somewhat predict the randomness of the random function. But I think in the context of this test, we can dismiss it. 

Other than that, looks great. Thanks for doing this!

trietlu (approved):
lgtm

PR #376, opened 2021-03-19
GTT-1099: Create CheckData component

miguelpdiaz8:
## Description

Create CheckData component. This PR only includes create the CheckData component and call it from AddChart. Wiring the changes to call the API and the refactoring will be part of another PRs.

## Testing

Added unit tests. You can test it here: https://migdizn.badger.wwps.aws.dev/admin/dashboard/ab13ccfa-17e5-4f5f-826d-0c628a621edc/add-chart

---


ferdingler:
Looks great! Although the `Hide from visualization` option should not hide the column in the table. It is just a configuration so that the column is hidden in Step 3 of the actual visualization. But in the `CheckData` step you still want to see all the columns in the dataset.

frontend/src/components/__tests__/CheckData.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+      onCancel={() => {}}
+      setSelectedHeaders={() => {}}
+      setHiddenColumns={() => {}}
+      selectedHeaders={new Set<string>()}
+      hiddenColumns={new Set<string>()}
+      register={() => {}}
+    />
+  );
+  expect(wrapper.container).toMatchSnapshot();
+});
```
ferdingler: the title of the unit test is not accurate. Also this snapshot test is not that useful because it renders an empty component without data. If you want to keep this as snapshot test, at least pass some data to the component so that it renders a table in my opinion.

frontend/src/components/CheckData.tsx
```diff
@@ -0,0 +1,165 @@
...
+                type="checkbox"
+                name={header}
+                defaultChecked={props.selectedHeaders.has(header)}
+                onChange={handleSelectedHeadersChange}
+                ref={props.register()}
+              />
+              <label
+                className="usa-checkbox__label"
+                htmlFor={`checkbox-header-${i}`}
+              ></label>
```
ferdingler, line 72: The content of this label is empty, I suggest you should add something so that screen readers can understand what his checkbox is for. For example: 

```jsx
<label className="sr-only"> Select column {i} </label>
```

miguelpdiaz8: I tried it but it is not working.

frontend/src/components/__tests__/CheckData.test.tsx
miguelpdiaz8: Fixed

miguelpdiaz8:
> Looks great! Although the `Hide from visualization` option should not hide the column in the table. It is just a configuration so that the column is hidden in Step 3 of the actual visualization. But in the `CheckData` step you still want to see all the columns in the dataset.

Done

ferdingler (approved):

PR #377, opened 2021-03-22
GTT-1120 Added full screen preview

jenboydclark:
## Description

Added the option to make the preview of a dashboard widget full screen on the add and edit pages for all 5 widget types.

## Testing

Added unit tests for the widget types that have mocked preview content.
See it in action here --> https://dvinr4n9hngnv.cloudfront.net/admin

---

### DIFF
--- a/frontend/src/hooks/dashboard-preview-hooks.tsx
+++ b/frontend/src/hooks/dashboard-preview-hooks.tsx
@@ -1,0 +1,26 @@
+import React, { useState } from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faPlusSquare, faMinusSquare } from "@fortawesome/free-solid-svg-icons";
+import Button from "../components/Button";
+
+export function useFullPreview() {
+  const [fullPreview, setFullPreview] = useState(false);
+
+  const fullPreviewToggle = () => {
+    setFullPreview(!fullPreview);
+  };
+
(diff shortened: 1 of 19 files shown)


frontend/src/hooks/dashboard-preview-hooks.tsx
```diff
@@ -0,0 +1,26 @@
...
+    </Button>
+  ) : (
+    <Button onClick={fullPreviewToggle} variant="outline" type="button">
+      <FontAwesomeIcon icon={faPlusSquare} className={"margin-right-1"} />
+      Expand preview
+    </Button>
+  );
+
+  return { fullPreviewToggle, fullPreviewButton, fullPreview };
+}
```
ferdingler, line 26: Interesting implementation, a hook that returns HTML 🤔 . Never seen this before. I'm pretty sure that hooks were not meant to return components. But I can't think of why this would be a bad idea at the moment either. 

Did you run into performance issues by any chance? Any lagginess in the UI? I'm thinking of a situation where the hook is always returning a new HTML value and thus causing extra renders.

jenboydclark: Nope, I didn't run into any performance issues.

ferdingler: Ok, let's leave it then and keep an eye on it :)

ferdingler (approved):

PR #383, opened 2021-03-25
Backend - Add significantDigitLabels property to Chart, Table and Metrics

ferdingler:
## Description

Added new property `significantDigitLabels` in Chart, Table and Metrics widgets. Refactored the `fromItem` function in WidgetFactory to support defining default values for properties that don't exist in the item in dynamodb. This will happen more often as we add new properties to our models that won't be present in existing old items. 

## Testing

Ran locally and verified with postman that the new property is being saved and retrieved. 

---


trietlu (approved):
lgtm

PR #396, opened 2021-03-30
GTT-1165: Added error messaging when adding content items

ajmokotoff:
## Description

Added disabled tooltip to the button component. Then using this, I met the criteria from GTT-1012.

## Testing

Ran tests and tested locally.

---


ajmokotoff:
Regarding file: frontend/src/containers/__tests__/AddMetrics.test.tsx I was unable to make the original test work (in a reasonable amount of time) of ensuring the backend was called upon metric upload. This is because I added in enforcement that ensures at least one metric is added to the metrics before it is allowed to get uploaded. In the mean time, I modified the test, to ensure that the backend was not called, if you did not add a metric. But I need help mocking the add metrics part.

frontend/src/components/VisualizeTable.tsx
```diff
@@ -85,6 +86,14 @@ function VisualizeTable(props: Props) {
...
     <div className="grid-row width-desktop">
       <div className="grid-col-5" hidden={props.fullPreview}>
+        {props.errors.title ? (
+          <Alert
+            type="error"
+            message="Resolve error(s) to add the table"
+          ></Alert>
+        ) : (
+          ""
+        )}
```
ferdingler: When there is no else condition, it's cleaner to simply write: 
```jsx
{props.error.title && ( <Alert ... /> )}
```

frontend/src/containers/AddMetrics.tsx
```diff
@@ -471,6 +485,18 @@ function AddMetrics() {
...
+                        type="error"
+                        message={
+                          submittedMetricsNum === 0
+                            ? "Enter at least one metric to continue"
+                            : "Resolve error(s) to add the text"
+                        }
+                      ></Alert>
+                    ) : (
+                      ""
+                    )}
```
ferdingler: Same comment as above.

ferdingler (approved):

PR #401, opened 2021-04-01
fix: Part-to-whole is broken on very small screens

ferdingler:
## Description

Making the height of the `Part to whole chart` be dynamic depending on the amount of parts in the chart and depending on the screen size. I wrote a hook to get the window width and heights as it gets resized. 

Waiting for Jason and Andrew to confirm that they are okay with this fix. 

## Testing

Verified that all part-to-whole charts in this dashboard are displayed correctly in a smartphone: https://fdingler.badger.wwps.aws.dev/homenick-and-sons . 

---


trietlu (approved):
lgtm

PR #413, opened 2021-04-06
GTT-955: Add option to show/hide numbers on bar and columns charts

miguelpdiaz8:
## Description

Add option to show/hide numbers on bar and columns charts

## Testing

---

### DIFF
--- a/frontend/src/components/ColumnChartWidget.tsx
+++ b/frontend/src/components/ColumnChartWidget.tsx
@@ -7,6 +7,7 @@ import {
   BarChart,
   Bar,
   Legend,
+  LabelList,
   ResponsiveContainer,
   CartesianGrid,
   Tooltip,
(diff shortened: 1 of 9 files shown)


frontend/src/components/ColumnChartWidget.tsx
```diff
@@ -82,6 +84,20 @@ const ColumnChartWidget = (props: Props) => {
     }
   };
 
+  const renderCustomizedLabel = (props: any) => {
+    const { x, y, width, value } = props;
+
+    return (
+      <text z={1} x={x + width / 2} y={y} dy={-6} textAnchor="middle">
```
trietlu, line 91: What does -6 mean ?

ferdingler (approved):

frontend/src/components/ColumnChartWidget.tsx
miguelpdiaz8: The label margin with the bar from inside, so if we pick a negative number, it will be shown outside and that is what we want to.

PR #417, opened 2021-04-07
GTT-1077 Access denied screen

ferdingler:
## Description

Created an access denied screen that the user is redirected to in the case where they don't have a role assigned. In order for this to work, I had to change the backend endpoint to fetch settings to avoid requiring a role. Any authenticated user will be able to fetch settings regardless of the role. Settings are necessary to load elements of the website. 

## Testing

Created a user in Cognito manually without assigning a role to it. Verified that logging in with that user takes me to the AccessDenied page. 

---


ajmokotoff (approved):
lgtm!

PR #437, opened 2021-04-15
Translated AddImageScreen and updated other AddScreens

ghost:
## Description

Update the AddImage Screen. 

## Testing

I ran the environment locally to make sure everything is right.

---

### DIFF
--- a/frontend/src/locales/en/translation.json
+++ b/frontend/src/locales/en/translation.json
@@ -538,7 +538,6 @@
     "CheckData": "Check data",
     "Visualize": "Visualize"
   },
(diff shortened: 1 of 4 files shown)


frontend/src/locales/en/translation.json
```diff
@@ -538,50 +538,26 @@
     "CheckData": "Check data",
     "Visualize": "Visualize"
   },
-  "Bar": "Bar",
```
miguelpdiaz8: We should not delete these. That is causing a lot of tests fail.

miguelpdiaz8 (approved):

PR #459, opened 2021-05-08
GTT-1224 Allow Live Preview to be Sticky

ghost:
## Description

In order to make the live previews sticky so that an editor can scroll down the options and still be able to see their live changes, I created a sticky-preview class and assigned it to the divs that contained the previews. 

Also, some of the divs inside the AddImage and EditImage containers had typos in their classNames so I corrected them.

## Testing

I updated the tests to include the new sticky-preview class. All tests passed. 

---

### DIFF
--- a/frontend/src/containers/AddImage.tsx
+++ b/frontend/src/containers/AddImage.tsx
@@ -285,7 +285,7 @@ function AddImage() {
             </form>
           </PrimaryActionBar>
         </div>
-        <div className={fullPreview ? "gril-col-12" : "grid-col-6"}>
+        <div className={fullPreview ? "grid-col-12" : "grid-col-6"}>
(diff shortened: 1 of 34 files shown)


frontend/src/containers/AddImage.tsx
```diff
@@ -285,7 +285,7 @@ function AddImage() {
             </form>
           </PrimaryActionBar>
         </div>
-        <div className={fullPreview ? "gril-col-12" : "grid-col-6"}>
+        <div className={fullPreview ? "grid-col-12" : "grid-col-6"}>
```
miguelpdiaz8, line 288: Shouldn't we add the sticky here for AddImage and AddMetrics?

miguelpdiaz8 (approved):

PR #471, opened 2021-05-20
Fix: Add (Copy) to widget name when duplicating

miguelpdiaz8:
## Description

Add (Copy) to widget name when duplicating.

## Testing

Tested it locally.

---


trietlu (commented):
Should this be localized? Ask Product if they want to always use "Copy"

trietlu (approved):
lgtm

PR #514, opened 2021-07-08
[BUG] Ensure logo file gets stored in the proper bucket

ajmokotoff:
## Description

Currently the EnvConfig lambda has an environment variable for the content bucket, but it is not used by the lambda function to actually include the contentBucket in the env.js file.

This is causing the current code to search and store the logo in the datasets bucket, since the content bucket cannot be resolved to an actual bucket (since its missing from the env.js).

## Testing

We modified the env.js file locally, and uploaded a new file, and ensure the logo was stored in the proper place.

---


yinxudeng (approved):
lgtm

miguelpdiaz8 (approved):

PR #520, opened 2021-07-13
GTT-1410: added IAM policy for s3 content bucket

yinxudeng:
## Description

Andrew and I noticed that there is no policy for s3 content bucket, so we made the change in cdk.

## Testing

All local tests and integrations tests passed. 

---


ajmokotoff (approved):
Approved! Yin and I noticed strange behavior, so we took a deep dive and noticed that the content bucket was missing from the S3 policy in the api handler. We went ahead and modified the policy.

miguelpdiaz8 (approved):

PR #523, opened 2021-07-14
GTT-1403: "Your logo" never loads once it's replaced

yinxudeng:
## Description

Made default logo hidden once users upload their own logo; always replaced old logo with new logo in the content S3 bucket when user uploads one 

Demo: https://d1p86h8o9mi768.cloudfront.net/

## Testing

Modified unit tests that call uploadLogo(). All unit tests and integration tests passed. 

---


frontend/src/components/Logo.tsx
```diff
@@ -7,6 +7,22 @@ function Logo() {
...
+    if (secondUpdate.current) {
+      if (firstUpdate.current) {
+        firstUpdate.current = false;
+        return;
+      }
+      secondUpdate.current = false;
+      return;
+    }
+    setToHide(false);
+  }, [loadingFile]);
```
ajmokotoff, line 26: If we change `loadingFile` to `logo`, we can prevent useEffect from running two times, and it will only run once. In this case, we can remove the complexity of having the secondUpdate useRef.

ajmokotoff (commented):
Excellent work Yin, this was a rather complex task that you solved. 

Two nitpicks:

1. I left a comment on simplifying the useEffect function
2. Can we externalize this logic so we can use it on your favicon task?

frontend/src/services/StorageService.ts
```diff
@@ -175,7 +175,19 @@ async function uploadImage(
...
 async function uploadLogo(rawFile: File): Promise<string> {
-  return await uploadImage(rawFile, "logo", EnvConfig.contentBucket);
+  const mimeType = rawFile.type;
+  const extension = imageFileTypes[mimeType as keyof ValidFileTypes];
+
+  if (!extension) {
+    throw new Error("File type is not supported");
+  }
+
+  const fileS3Key = "uploadedLogo".concat(extension);
```
ajmokotoff: This might not work if the extensions are different

frontend/src/services/StorageService.ts
```diff
@@ -175,7 +175,19 @@ async function uploadImage(
 }
 
 async function uploadLogo(rawFile: File): Promise<string> {
-  return await uploadImage(rawFile, "logo", EnvConfig.contentBucket);
```
ajmokotoff, line 178: Thoughts on having an optional parameter instead of having some duplicate code?

frontend/src/hooks/logo-hooks.ts
```diff
@@ -14,7 +14,7 @@ export function useLogo(s3Key?: string): UseLogoHook {
     if (s3Key) {
       try {
         setLoading(true);
-        const data = await StorageService.downloadLogo(s3Key);
+        const data = await StorageService.downloadLogo(s3Key, () => {});
```
ajmokotoff: Good find, I'm wondering how this was working before?

frontend/src/services/StorageService.ts
yinxudeng: Good catch.

frontend/src/services/StorageService.ts
yinxudeng: I will refactor this part of the code base.

frontend/src/hooks/logo-hooks.ts
yinxudeng: It seems like TypeScript (or JavaScript) function parameters are defaulted to undefined according to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters). In our case, the passed in function is called when downloadFile encounters an error. I'll change the dummy function to the translation function.

frontend/src/components/Logo.tsx
yinxudeng: useEffect cannot depend on logo because if the logo never loads (in other words we are using the default logo), we will never set toHide to false. That means the default logo will never show up. I do agree the logic here is not straight-forward enough.

miguelpdiaz8 (approved):

PR #525, opened 2021-07-15
GTT-1394: Updated Create Metric Dataset API to include symbols and fixed bug with currency in Add Metrics page

ghost:
## Description

Updated the "Create Metric Dataset" API to include symbols (like percentage and currency) and also fixed bug with currency in the Add Metrics page. Will submit a separate PR for the same functionality in the "Update Metric Dataset" API.

## Testing

Tested APIs on Postman, personal badger instance, created corresponding unit tests, ran ./test.sh, and ran prettier

---

### DIFF
--- a/backend/src/lib/controllers/ingestapi-ctrl.ts
+++ b/backend/src/lib/controllers/ingestapi-ctrl.ts
@@ -23,6 +23,43 @@ async function createDataset(req: Request, res: Response) {
     return res.status(400).send("Missing required field `data`");
   }
 
+  if (metadata.schema === "Metrics") {
+    for (let i = 0; i < data.length; i++) {
+      //symbol should be valid input or empty
+      if (!["", "Currency", "Percentage"].includes(data[i].percentage)) {
+        return res
+          .status(400)
+          .send(
+            "Invalid symbol type. Choose either `Currency`, `Percentage` or ``"
+          );
+      }
+      //currency should be valid input or empty
+      if (!["", "Dollar $", "Euro €", "Pound £"].includes(data[i].currency)) {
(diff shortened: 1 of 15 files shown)


backend/src/lib/controllers/ingestapi-ctrl.ts
```diff
@@ -23,6 +23,43 @@ async function createDataset(req: Request, res: Response) {
     return res.status(400).send("Missing required field `data`");
   }
 
+  if (metadata.schema === "Metrics") {
+    for (let i = 0; i < data.length; i++) {
+      //symbol should be valid input or empty
+      if (!["", "Currency", "Percentage"].includes(data[i].percentage)) {
```
trietlu: @AakashBalaji don't we have to localize "Currency" and "Percentage"

backend/src/lib/controllers/ingestapi-ctrl.ts
```diff
@@ -23,6 +23,43 @@ async function createDataset(req: Request, res: Response) {
     return res.status(400).send("Missing required field `data`");
   }
 
+  if (metadata.schema === "Metrics") {
+    for (let i = 0; i < data.length; i++) {
```
trietlu: is it better to use for..of loop, like for (let datum of data)

backend/src/lib/controllers/ingestapi-ctrl.ts
ghost: Thanks for letting me know Triet, just added localizations for currency and percentage

miguelpdiaz8 (approved):

PR #529, opened 2021-07-19
#GTT-1436: Fixed bug in auto-top positioning of content item page and used enum to represent percentage and currency

ghost:
## Description

Fixed bug in auto-top positioning of content item page and used enum to represent percentage and currency. Edited the Add/Edit Chart, Metrics and Table pages as well as the tickformatter.ts page.

## Testing

Tested changes on personal badger instance, ran ./test.sh, and prettier.
Check it out here: https://dhhjkug7z9sph.cloudfront.net/admin/dashboard/edit/0f83646b-77dd-4d2a-b264-bcf56bc4e7db

---

### DIFF
--- a/frontend/src/containers/AddChart.tsx
+++ b/frontend/src/containers/AddChart.tsx
@@ -14,6 +14,7 @@ import {
   useDatasets,
   useFullPreview,
   useChangeBackgroundColor,
+  useScrollUp,
 } from "../hooks";
 import StorageService from "../services/StorageService";
 import BackendService from "../services/BackendService";
@@ -393,6 +394,7 @@ function AddChart() {
   );
 
   useChangeBackgroundColor();
+  useScrollUp();
 
   return (
     <>
--- a/frontend/src/containers/AddChart.tsx
+++ b/frontend/src/containers/AddChart.tsx
@@ -103,7 +103,7 @@ function AddChart() {
   const [currencyTypes, setCurrencyTypes] = useState<
     Map<string, CurrencyDataType>
   >(new Map<string, CurrencyDataType>());
-  const [oldStep, setoldStep] = useState<number>(-1);
+  const [oldStep, setOldStep] = useState<number>(-1);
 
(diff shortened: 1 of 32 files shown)


frontend/src/containers/AddChart.tsx
```diff
@@ -103,6 +103,7 @@ function AddChart() {
   const [currencyTypes, setCurrencyTypes] = useState<
     Map<string, CurrencyDataType>
   >(new Map<string, CurrencyDataType>());
+  const [oldStep, setoldStep] = useState<number>(-1);
```
yinxudeng: Should we name the setter setOldStep since TypsScript uses camelCase? This also happens in other files below.

frontend/src/hooks/scrollup-hooks.ts
```diff
@@ -1,7 +1,10 @@
 import { useEffect } from "react";
 
-export function useScrollUp(): any {
+export function useScrollUp(oldStep: number, step: number): any {
   useEffect(() => {
-    window.scrollTo(0, 0);
-  });
+    if (oldStep !== step) {
+      window.scrollTo(0, 0);
+      oldStep = step;
```
yinxudeng: Should we also pass in setOldStep to this custom hook so that the state setter can be called instead assigning a value to the state?

frontend/src/containers/AddChart.tsx
ghost: yes, good catch!

frontend/src/hooks/__mocks__/index.tsx
```diff
@@ -561,4 +561,8 @@ export function useWindowSize() {
 }
 
 export function useChangeBackgroundColor() {}
-export function useScrollUp() {}
+export function useScrollUp(
+  oldstep: number,
```
yinxudeng: Should we have oldStep here?

yinxudeng (approved):

PR #561, opened 2021-08-07
Create new section content item and drag and drop for sections

miguelpdiaz8:
## Description

Create new section content item and drag and drop for sections.

Here is the code to create a section content item. Both reordering are implemented, I tested the arrow feature a lot but the drag and drop is not heavily tested and I would expect due to complexity that we are missing some edges cases. Please test it thoroughly before merging. As you see I had also to update the version of prettier we run on github because I was getting some unrelated issues, now we are using always the last version.

## Testing

Added unit tests. You can test it here: https://migdizn.badger.wwps.aws.dev/admin/dashboard/edit/f27e2bca-f960-4af8-92e9-8a31b20db317

---


yinxudeng:
After some initial testing, here are a few edge cases our current implementation is missing:
1. Clicking the up arrow or down arrow of a non-section content item sometimes move itself inside/outside of a section content item for half a second before moving itself back to the original position.
2. If there is a section content item at the very end, arrow and drag-and-drop cannot move the non-section content items inside to underneath. 
3. Drag-and-drop section content item can sometimes move the non-section content items inside to the outside or change other sections' content items, but I haven't found a way to reproduce this consistently.

trietlu (approved):
lgtm

PR #562, opened 2021-08-10
GTT-1516: [Bug] Fixed missing labels in bar chart when there is too much data

yinxudeng:
## Description

Modify the height of bar chart to grow dynamically with the number of data.

## Testing

Passed all unit tests and integration tests. You can check it out here: https://d1p86h8o9mi768.cloudfront.net/bar-chart-missing-label-bug

---


trietlu (changes requested):
@yinxudeng I know you didn't make the original code change, but can we comment and/or use constants for what the numbers 800, 400, and 15 mean

yinxudeng:
> @yinxudeng I know you didn't make the original code change, but can we comment and/or use constants for what the numbers 800, 400, and 15 mean

Thanks for catching this! I refactored code and added comments similar to the following:
400px is enough for datasets with <= 15 rows and <= 3 columns, and this should cover most common use cases.
When datasets have >15 rows and >3 columns, it would require additional height increments.

trietlu (approved):
lgtm

PR #585, opened 2021-08-25
Gtt 1538

PhilTower:
## Description
Updated CDK to 1.119.0 and cdk-lambda-at-edge-pattern to 1.39.0
This resolves the issue with HttpHeader not getting a unique name with deploys. I updated all the packages to resolve typescript version mismatch.

This is has changes for GTT-1539 which is needed to make the package update work.
Added:
``` 
 targetKmsKey.addAlias(`PDoA/OpsNotifications-${props.environment}`);
```

For multiple deploys the alias needs unique per deploy.  This will be unique per environment not per deploy. Right now there are other constraints with multiple stack deploys using cloud formation.

## Testing

I have tested by deploying an instance of performance dashboard using the link in the ReadMe. Then I made multiple deployments using the deploy.sh script since it allows the environment to be set.
Then in each deployment I logged into the admin site. 

---


trietlu (approved):
lgtm

PR #589, opened 2021-08-30
GTT-1010 Send user agent string can SDK call to dynamodb

trietlu:
## Description

**GTT-1010** As a Solutions release, we want to add a user-agent string to AWS SDK calls so we have metrics on service usage.  The only AWS SDK calls the backend makes is Cognito and DynamoDB (other SDK calls are through Amplify).  We only added the user-agent string to the DynamoDB calls, as calls to Cognito are few.  In the backend code, we import package.json to get the version number to use with the user-agent string.  That causes typescript to change the folder structure of the backend build to become build/src/ instead of build/.  We thus had to change the lambda construct in CDK to change the path of the handler to src/lambda/, from lambda/

**GTT-1544** Update the version to 1.0.6, along with the documentation

## Testing

* unit tests
* integration tests in postman
* e2e tests
* manual tests

---


miguelpdiaz8 (approved):
LGTM

PR #594, opened 2021-09-01
Gtt 1476

PhilTower:
## Description

**E2E Changes**
I upped the width and height of the view portal on the e2e tests since several of the failures were due to the element being off screen.

**App Changes**
Added AUTH_REGION environment variable. This is required because now the app can be deployed to one region but the Cognito service it uses is in another region. Region needs to be specified in order for the user pool to be found. 

**Deploy Script Changes**
I changed the deploy script to hide the changes of same vs split region deploy. 
Deploying to the same region is unchanged `./deploy.sh perfdash1`
Deploying to a region with cognito in a different region `./deploy.sh perfash1 us-west-1`
The deploy.sh was moved to deployscripts/oneregion.sh

**CDK Changes**
I tired making this change in a few different ways. Region is supposed to be set able in typescript, either once the region is set the value is immutable or there is a bug because the other region was not being used. I ran into some issues with template generation because of the HTTP Headers library not handling CF parameters in the stack. 

The change I made was this, in the deploy script save the current AWS region to a variable, set AWS_REGION to the passed region. Deploy the AuthStack and save the variables to a JSON file. Parse the JSON file using grep and grab the values needed for the other stacks. I know there is a JSON parser CLI tools for bash but everyone should have grep. "Deploy" the parameters using the AuthParamsStack, which gets the required values in and avoids the issues with the HttpHeader library. 

## Testing

I deployed two multi region and to single region and ran e2e tests on both.
Running the e2e uncovered the issue, with failing to create users. That is why I needed to add the AUTH_REGION environment variable and pass it to the Cognito service.

---


trietlu:
We will only merge this PR after the next round of AppSec review.  Marking as draft until then

PR #601, opened 2021-09-07
Gtt 1535

PhilTower:
## Description

Changed how the column meta data was selected to map from index. Depending on how many columns are in the data set and how many columns have meta data, this may or may not be the correct index.

To resolve that I changed the lookup from index based to using the name of the bar/column to get the correct meta data.

## Testing

Added a integration test chart-render.spec.ts to verify the symbols show up in both column and bar charts 

---


trietlu (approved):
lgtm

PR #604, opened 2021-09-07
Gtt 1530

PhilTower:
## Description

Added the option for display percentages, which calculates the columns percentage based on the data.
This was added for donut and pie charts.
Removed some repeated formatting code by formatting the labels once on useMemo and then selecting the appropriate display value.

## Still Need

Final test for Display percentage options
Translations for that option

## Testing

Manually tested that the labels appeared correctly as described in the AC of the task.
Added unit tests for a ChartDataFormatters which removed some logic from the components for clarity.
Added 1 e2e test for the pie chart which only covers 1 permutation of options.

---


frontend/src/components/ChartWidget.tsx
```diff
@@ -109,6 +109,7 @@ function ChartWidgetComponent(props: Props) {
           parts={keys}
           data={filteredJson}
           significantDigitLabels={content.significantDigitLabels}
+          computePercentages={content.computePercentages}
```
miguelpdiaz8: Why to add this here if we did not change the PartToWhole component?

frontend/src/containers/AddChart.tsx
```diff
@@ -190,6 +192,10 @@ function AddChart() {
             values.chartType === ChartType.DonutChart) && {
             dataLabels: values.dataLabels,
           }),
+          ...((values.chartType === ChartType.PieChart ||
```
miguelpdiaz8, line 195: Same here, is PartToWhole part of this PR?

frontend/src/containers/EditChart.tsx
```diff
@@ -381,6 +385,10 @@ function EditChart() {
             values.chartType === ChartType.DonutChart) && {
             dataLabels: values.dataLabels,
           }),
+          ...((values.chartType === ChartType.PieChart ||
```
miguelpdiaz8, line 388: Same here, is PartToWhole part of this PR?

frontend/src/locales/es/translation.json
```diff
@@ -168,6 +168,7 @@
     "ChartTypeError": "Seleccione un tipo de gráfico",
     "DisplayHorizontalScroll": "Mostrar con desplazamiento horizontal",
     "ShowDataLabels": "Mostrar etiquetas de datos",
+    "ComputePercentages": "!!NEED TRANSLATION!!",
```
miguelpdiaz8: We prefer not to change the es and pt translation files, only the English one, that way is easier for us later.

frontend/src/locales/pt/translation.json
```diff
@@ -168,6 +168,7 @@
     "ChartTypeError": "Selecione um tipo de gráfico",
     "DisplayHorizontalScroll": "Exibir com rolagem horizontal",
     "ShowDataLabels": "Mostrar rótulos de dados",
+    "ComputePercentages": "!!NEED TRANSLATION!!",
```
miguelpdiaz8: We prefer not to change the es and pt translation files, only the English one, that way is easier for us later.

miguelpdiaz8 (commented):
Added some comments, we also need you to deploy these changes to your personal AWS account and create users for us to test the actual functionality, thanks.

frontend/src/components/ChartWidget.tsx
PhilTower: Missed that when I was removing partofwhole

frontend/src/containers/AddChart.tsx
PhilTower: No, I missed one reference that I should have deleted

frontend/src/locales/es/translation.json
PhilTower: Ok

frontend/src/containers/EditChart.tsx
PhilTower: see above

PhilTower:
@miguelpdiaz8 

Here is the location of my personal instance: https://d2srrxle2f48hl.cloudfront.net/
added you, triet and jason

frontend/src/services/ChartDataFormatter.ts
```diff
@@ -0,0 +1,90 @@
+import { isString } from "util";
```
miguelpdiaz8: isString is deprecated, please use: typeof value === 'string' instead. I think in this case we can delete this line because it is not being used anyway.

miguelpdiaz8 (approved):
LGTM

PR #607, opened 2021-09-09
GTT-1537: Significant data labels need to be applied when currency is used

miguelpdiaz8:
## Description

Significant data labels need to be applied when currency is used.

## Testing

Tested it locally. Updated tests.
---


trietlu (approved):
lgtm

PR #608, opened 2021-09-10
Fix filtered json logic

mmarkelov:
## Description
In JS `filteredRow !== {}` does not have any sense, cause it will be always **true**. One of the way to check if the object is empty is usage of its keys length: `Object.keys(filteredRow).length` 

---


trietlu:
Hello @mmarkelov, thanks for making this pull request.  As per the [Contributing](https://github.com/awslabs/performance-dashboard-on-aws/blob/mainline/CONTRIBUTING.md#contributing-via-pull-requests) guideline, please make your pull request against the mainline-dev branch.

PR #641, opened 2021-10-13
GTT-1573: Fix position of actions dropdown on mobile

miguelpdiaz8:
## Description

Fix position of actions dropdown on mobile.

## Testing

Updated unit tests.

---


frontend/src/components/DataTable.tsx
```diff
@@ -60,7 +67,11 @@ function DataTable({ rows, columns, columnsMetadata, fileName }: Props) {
 
   return (
     <>
-      <div className="text-right">
+      <div
+        className={
+          showMobilePreview ? "margin-left-05" : "text-right margin-right-05"
```
mabreuortega, line 72: We should start moving all this custom styling to css files. It is hard to mantain in the long run if we have the styling all over the place.
Something like

```
div[showMobilePreview=true] {
  ...
}
```

In this case this will not apply cuz you are switching classes and there is not inheritance in CSS so far but it is something we need to look forward in other places

mabreuortega (approved):
LGTM

PR #647, opened 2021-10-14
GTT-1581: Add accessible labels to the Check Data column checkboxes

chasechow7:
## Description

Added accessibility labels to the checkboxes so the text-to-speech screen reader will output the column header name

## Testing

- Updated and ran Unit Tests
- Checked functionality with Screen Reader extension

---


trietlu:
@chasechow7 how are we testing that the aria label is present?

chasechow7:
> @chasechow7 how are we testing that the aria label is present?

I checked the aria-label is present under the Elements tab when inspecting the page
image

miguelpdiaz8:
They are also present in the snapshot tests

trietlu:
Bad Triet, read it too fast :-)

trietlu (approved):
lgtm

PR #648, opened 2021-10-14
GTT-1581: Add accessible labels to the Check Data column checkboxes

chasechow7:
## Description

Added accessibility labels to the checkboxes so the text-to-speech screen reader will output the column header name

## Testing

- Updated and ran Unit Tests
- Checked functionality with Screen Reader extension

---


trietlu (approved):
lgtm

PR #663, opened 2021-10-25
GTT-1590 - Removed 403 error handler.  Added bucket policy to allow for CF to access bucket

trietlu:
## Description

When we upgraded CDK, there was a breaking change and we added a 403 error handler based on the CDK ticket below

https://github.com/aws/aws-cdk/issues/13272 

However, the better fix is to do what the CDK author pointed out, to give the Cloudfront identity access to the S3 bucket.  

bucket.grantRead(originAccessIdentity);

The 403 error handler we added prevented us from using the WAF, which would fail block request with a 403

The other change is to the sample CFT to create a WAF that will block access based on origin CIDR range and the path that the request is accessing.  For example, block all access that didn't originate from 10.0.0.0/8 and has /admin in the path.

## Testing

Installed in my AWS account and tested

---

### DIFF
--- a/tools/WAF-enterprise-only.json
+++ b/tools/WAF-enterprise-only.json
@@ -3,11 +3,39 @@
     "Parameters": {
       "InternalCIDR": {
         "Type": "CommaDelimitedList"
-      }
-    },  
+      },
+      "RestrictPublicDashboard": {
+        "Description": "Whether to restrict access to public dashboards.",
+        "Default": false,
+        "Type": "String",
+        "AllowedValues": [
+            true,
+            false
+        ]
+      } 
+    },
+    "Conditions": {
+      "RestrictPublicDashboardToEnterprise": {
+          "Fn::Equals": [
+              true,
+              {
+                  "Ref": "RestrictPublicDashboard"
+              }
+          ]
+      },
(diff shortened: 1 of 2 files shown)


tools/WAF-enterprise-only.json
```diff
@@ -101,47 +114,34 @@
...
+                          {
+                            "Type": "NONE",
+                            "Priority": 0
+                          }
+                        ]
+                      }
                     }
                   ]
-                }
+                }
```
chasechow7: Seems to be some empty white space. Surprised prettier doesn't catch this.

tools/WAF-enterprise-only.json
```diff
@@ -101,47 +114,34 @@
...
-              },
-              "VisibilityConfig": {
-                "CloudWatchMetricsEnabled": true,
-                "MetricName": "PublicPage",
-                "SampledRequestsEnabled": true
+              "Action": {
+                "Block": {}
               }
-            }
+            }
```
chasechow7: some empty white spaces

chasechow7 (changes requested):

chasechow7 (approved):
looks good!

PR #673, opened 2021-10-29
Dashboard Listings Tab Accessibility

chasechow7:
## Description

Add accessibility functionality to dashboard listings tabs

- Ability to tab to core tabs (Drafts, Publish Queue, etc.)
- Add keyboard functionality: Ability to select tab using 'Enter'
- Aria Label for screen reader will read: "Dashboard Tab {Name}"
- Aria Label will indicate currently selected tab with "Active Dashboard Tab {Name}"

## Testing
Can test it here: https://d2wo27gatac4jo.cloudfront.net/admin/dashboards

- Turn on screen reader and tab through the core dashboard tabs
- Press Enter to select highlighted tab
- When tabbing to active tab. Can confirm Aria Label indicates this is the active tab

-Added and passed Unit Testing

---


trietlu (commented):
@chasechow7 I did some testing and it looks good

@miguelpdiaz8 Is this the only tab component where we need this change? The section tab doesn't share code with this component right?

chasechow7:
> @chasechow7 I did some testing and it looks good
> 
> @miguelpdiaz8 Is this the only tab component where we need this change? The section tab doesn't share code with this component right?

Ohh... Good call. I think the section tab does share code with this. Instead of hard coding in "dashboard" in the aria label. I'll make sure its generated dynamically so it'll say "Section Tab {name}" when used in section

miguelpdiaz8:
> > @chasechow7 I did some testing and it looks good
> > @miguelpdiaz8 Is this the only tab component where we need this change? The section tab doesn't share code with this component right?
> 
> Ohh... Good call. I think the section tab does share code with this. Instead of hard coding in "dashboard" in the aria label. I'll make sure its generated dynamically so it'll say "Section Tab {name}" when used in section

You also need to add aria-labels for VerticalTabs and VerticalTab components

chasechow7:
> > > @chasechow7 I did some testing and it looks good
> > > @miguelpdiaz8 Is this the only tab component where we need this change? The section tab doesn't share code with this component right?
> > 
> > 
> > Ohh... Good call. I think the section tab does share code with this. Instead of hard coding in "dashboard" in the aria label. I'll make sure its generated dynamically so it'll say "Section Tab {name}" when used in section
> 
> You also need to add aria-labels for VerticalTabs and VerticalTab components

Hmm... Should that be a separate story if we're adding aria labels to the vertical tabs for the Section Widget?

chasechow7:
- Changed out strings to support Translation
- Added variable for container name to support either Section Widget or Dashboard Listing
- Updated Unit Tests

miguelpdiaz8 (approved):
LGTM

PR #674, opened 2021-11-01
GTT-1594: Make table pagination responsive base on sizing

miguelpdiaz8:
## Description

Make table pagination responsive base on sizing.

## Testing

Updated unit tests. You can test it here: https://migdizn.badger.wwps.aws.dev/admin/dashboards

---


chasechow7 (commented):
If we're removing scroll top from all the instances. Should the ScrollTop.tsx component be removed as well? Or, are we keeping it in case we'll use it again in the future?

miguelpdiaz8:
> If we're removing scroll top from all the instances. Should the ScrollTop.tsx component be removed as well? Or, are we keeping it in case we'll use it again in the future?

Yes, this component can be used again in the future in other parts of the app.

chasechow7 (approved):
looks good!

PR #678, opened 2021-11-05
Adds frontend support for searching inside published dashboards.

kharbutli-meta:
## Description

GTT-1557: Adds frontend support for searching inside of published dashboards. Given a query, searches text fields inside published dashboards, and displays those dashboards with content matching the query. Text snippets are also displayed.

## Testing

- Added unit tests and tested with it.
- Tested by running locally.

---

### DIFF
--- a/frontend/src/containers/Home.tsx
+++ b/frontend/src/containers/Home.tsx
@@ -16,7 +16,7 @@ function Home() {
   const dateFormatter = useDateTimeFormatter();
 
   const onSearch = (query: string) => {
-    window.location.assign("/search/"+query);
+    window.location.assign("/search/" + query);
   };
 
   const onClear = () => {
--- a/frontend/src/hooks/homepage-hooks.ts
+++ b/frontend/src/hooks/homepage-hooks.ts
@@ -65,8 +65,9 @@ type UsePublicHomepageSearchHook = {
   homepage: PublicHomepage;
 };
 
-export function usePublicHomepageSearch(query :string)
-    : UsePublicHomepageSearchHook {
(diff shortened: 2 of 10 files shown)


frontend/src/containers/Home.tsx
```diff
@@ -67,8 +54,8 @@ function Home() {
             onSubmit={onSearch}
             size="big"
             onClear={onClear}
-            query={filter}
```
mabreuortega, line 70: looks like filter is not used anymore? If this is the case, you may want to clean up the state declaration.

frontend/src/containers/Home.tsx
```diff
@@ -67,8 +54,8 @@ function Home() {
...
-            results={filterPublicDashboards(homepage.dashboards).length}
+            query=""
```
mabreuortega, line 57: What does query "" means for the app?

frontend/src/hooks/homepage-hooks.ts
```diff
@@ -59,3 +59,34 @@ export function usePublicHomepage(): UsePublicHomepageHook {
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setLoading(true);
+      const data = await BackendService.fetchPublicHomepageWithQuery(query);
+      setHomepage(data);
+      setLoading(false);
+    };
+    fetchData();
+  }, []);
```
mabreuortega, line 86: There is no dependency here, which means that this code is executed only once. Is that intented?

mabreuortega (commented):
Can you deploy the current version to your local and invite us there to check the functionality?

PR #744, opened 2022-01-04
Adds 'copy' to new 'Actions' drop-down menus

kharbutli-meta:
## Description

GTT-1788: Creates drop-down menus in the Admin's view of a published or archived dashboard and adds the copy functionality in it.

## Testing

* Unit tests: Updated.
* Ran locally.

---


frontend/src/containers/__tests__/ViewDashboardAdmin.test.tsx
```diff
@@ -100,7 +100,7 @@ describe("dashboard published", () => {
   });
 
   test("renders an archive button", async () => {
-    expect(screen.getByRole("button", { name: "Archive" })).toBeInTheDocument();
+    expect(screen.getByRole("button", { name: "Actions" })).toBeInTheDocument();
```
mabreuortega, line 103: I think we should look for the Actions > Archive directly. Otherwise those 2 tests are redundant.

frontend/src/containers/ViewDashboardAdmin.tsx
```diff
@@ -513,6 +556,18 @@ function ViewDashboardAdmin() {
         buttonType={t("PreparePublishingModalButton")}
         buttonAction={onPublishDashboard}
       />
+
+      <Modal
+        isOpen={isOpenCopyModal}
+        closeModal={closeCopyModal}
+        title={`${t("DashboardListing.Copy")} "${dashboard.name}"?`}
+        message={`${t("DashboardListing.ThisWillCopy")}
```
mabreuortega: Can we refactor this to properly use the i18n formatter.

translation.json
```
"message": "This will create a draft copy of this dashboard",
"message_plural": "This will create a draft copy of these dashboards",
```

and use it like

```
message={t("message", { count: selectedList.length }}
```

frontend/src/locales/en/translation.json
```diff
@@ -725,7 +725,7 @@
     "Published": "Published",
     "Archived": "Archived",
     "Copy": "Copy",
-    "ThisWillCopy": "This will copy",
+    "ThisWillCopy": "This will create a draft copy of",
```
mabreuortega: This is shared with another copy actions in a different page but the structure is not the same. Will be recommended create another entry in the json file or a component with a unique translation that can be referenced in both places. My concern is that context of the two occurrences may change in the future.

mabreuortega (commented):
Comments added

frontend/src/locales/en/translation.json
kharbutli-meta: This change is based on Jason's request. He wants the statement "This will create a draft copy of" to be used when asking for confirmation to copy a dashboard whatever the status of the original dashboard is. I do not see a need to have two entries for the same statement. If this changes in the future, we can add a new statement entry then.

frontend/src/containers/ViewDashboardAdmin.tsx
kharbutli-meta: I see your point here, but I believe that keeping translation files entries short and simple can help ensure they are reusable in multiple code locations. The entry DashboardListing.ThisDashboard is used in multiple locations. I prefer reusing it along with its plural counterpart over creating two specific entries just for copying (one for singular and one for plural). Thoughts?

trietlu: @mazenkh-amazon, @mabreuortega Both points are valid and have merits.  In this case, let's go with Miguel's suggestion to use the  i18n plural formatter.  Plus, I think Copy should have its own specific message.  Reusing messages across multiple locations is good, but may lead to unintentional coupling, especially if the context for each are different.

frontend/src/containers/__tests__/ViewDashboardAdmin.test.tsx
mabreuortega: Do you want to remove one of the buttons tests? They are validating the same thing.

frontend/src/containers/ViewDashboardAdmin.tsx
```diff
@@ -513,6 +556,16 @@ function ViewDashboardAdmin() {
         buttonType={t("PreparePublishingModalButton")}
         buttonAction={onPublishDashboard}
       />
+
+      <Modal
+        isOpen={isOpenCopyModal}
+        closeModal={closeCopyModal}
+        title={`${t("DashboardListing.Copy")} "${dashboard.name}"?`}
```
mabreuortega: Now that we are going over translations will be nice if we can refactor this to use i18n interpolation

trasnlation.json
```
"Copy":  "Copy {dashboardName}"
``` 

usage
```
title={t("Copy", { dashboardName: dashboard.name })}
```

frontend/src/containers/ViewDashboardAdmin.tsx
```diff
@@ -109,6 +113,35 @@ function ViewDashboardAdmin() {
...
+    closeCopyModal();
+
+    if (dashboard) {
+      await BackendService.copyDashboard(dashboard.id);
+
+      history.push("/admin/dashboards?tab=draft", {
+        alert: {
+          type: "success",
+          message: `${dashboard.name}
+                    ${t("DashboardListing.DashboardSingular")}
```
mabreuortega: This message also can use some TLC regarding i18n usage of interpolation and plurals.

mabreuortega (commented):
It looking better but we are still using the same translation of the other page.

kharbutli-meta:
* Which two lines in the test are redundant. I cannot find the redundancy. Can you point out the line numbers?

mabreuortega:
> 

103, 139 https://github.com/awslabs/performance-dashboard-on-aws/pull/744/files/1dd35d5d02beade69f6fe7cb80d1372b091f55a8#diff-b6bc1dccac77f3b2ad8ec6197a559c54325efd02b007ce263c5effd53fa84513

kharbutli-meta:
Lines 103 and 139 cover two different tabs (published dashboard and archived dashboard), and therefore, two different code paths. It is necessary to keep both tests in case the code in one of the paths changes. Otherwise, we may fail to catch a bug if someone changes the code in the untested path.

mabreuortega:
> Lines 103 and 139 cover two different tabs (published dashboard and archived dashboard), and therefore, two different code paths. It is necessary to keep both tests in case the code in one of the paths changes. Otherwise, we may fail to catch a bug if someone changes the code in the untested path.

Ohhh, got it, thx for the update

mabreuortega:
> It looking better but we are still using the same translation of the other page.

@mazenkh-amazon what about this?

frontend/src/containers/ViewDashboardAdmin.tsx
mabreuortega: @mazenkh-amazon ^^^^

frontend/src/containers/ViewDashboardAdmin.tsx
mabreuortega: @mazenkh-amazon DashboardListing.

frontend/src/containers/ViewDashboardAdmin.tsx
```diff
@@ -109,6 +113,35 @@ function ViewDashboardAdmin() {
...
+                    ${t("DashboardListing.SuccessfullyCopied")}`,
```
mabreuortega: @mazenkh-amazon DashboardListing.

frontend/src/containers/ViewDashboardAdmin.tsx
mabreuortega: @mazenkh-amazon DashboardListing.

mabreuortega (approved):
LGTM

PR #751, opened 2022-01-13
Adds support for scaling image sizes when added/edited

kharbutli-meta:
## Description

Adds support, in both the backend and frontend, for scaling image sizes when added/edited.

## Testing

* Unit tests: Updated.
* Ran locally.

---


frontend/src/containers/AddImage.tsx
```diff
@@ -217,6 +225,74 @@ function AddImage() {
...
+                          label: t("AddImageScreen.25"),
+                        },
+                      ]}
+                    />
+                  </div>
+                  <div className="grid-col-3 margin-top-10">
+                    <div style={{ position: "absolute", bottom: "0px" }}>
+                      {scalePct === "100%" && (
+                        <img
+                          src={`${process.env.PUBLIC_URL}/images/image-scale-100.svg`}
```
mabreuortega: recomendation: refactor URL resolution to a method and create only one image component

```
<img
    src={getScaleUrl(scalePct)}
    width="100%"
    height="auto"
  />
```

frontend/src/components/ImageWidget.tsx
```diff
@@ -7,10 +7,11 @@ type Props = {
   file: File | undefined;
   summaryBelow: boolean;
   altText: string;
+  scalePct: string;
```
mabreuortega, line 10: recommendation: consider restricting type to allow only possible values

```
scalePct: "auto" | "25%" | "50%" | "75%" | "100%"
```

or use an enum type

```
export enum ScalePct {
   Auto: "auto"
...
}
```

frontend/src/locales/en/translation.json
```diff
@@ -896,7 +896,14 @@
     "ToggleSummary": "Show description below image",
     "Back": "Back",
     "DisabledToolTip": "Upload a file to continue",
-    "Cancel": "Cancel"
+    "Cancel": "Cancel",
+    "ImageSize": "Image size (optional)",
+    "ImageSizeHint": "How your image will display in the dashboard.",
+    "AsUploaded": "As uploaded (do not scale)",
+    "100": "100%",
```
mabreuortega: suggestion: use interpolation

```
"percentage": "{percent}%"
```

and use

```
t("percentage", { percent: scalePct })
```

mabreuortega: this percentage can be global and you can reuse across the app. x% have no changing context

mabreuortega (approved):
LGTM, multiple suggestion made.

frontend/src/containers/AddImage.tsx
kharbutli-meta: Good recommendation. I changed it to use an inline function.

frontend/src/components/ImageWidget.tsx
kharbutli-meta: I prefer to keep it as is for two reasons:
1. Makes it easier to add values in the future especially if we decide to allow a user to enter any scale value.
2. Makes other parts of the code simpler as we can insert the scale string value immediately as is.

frontend/src/locales/en/translation.json
kharbutli-meta: Done.

PR #764, opened 2022-01-26
Removes the aria-hidden attribute from the file selection field

kharbutli-meta:
## Description

To support accessibility, removes the aria-hidden attribute from the file selection field so that it can be accessed by assistive technology.

## Testing

* Unit tests.
* Local run using a screen reader.

---


trietlu (approved):
lgtm

PR #783, opened 2022-02-23
Improves the UX for the CheckData component

kharbutli-meta:
## Description

GTT-1701: Improves the UX for the Check Data step when adding/editing a table or chart.

Screen Shot 2022-02-23 at 11 03 02 AM

## Testing

* Unit tests.
* Local run.

---


trietlu (approved):
lgtm

mabreuortega:
Can we layout the formatting like the table? I have mixed feelings about the table to show the data and a flat list to style each row. I think the formatting settings may be presented with a table similar to the table showing the data or we may add a new row with the formatting data.

mabreuortega (changes requested):
Table footer width changes depending on selection.

image

image

mabreuortega:
another issue, go to https://d20ol8pdcmowfi.cloudfront.net/admin/dashboard/2668bb2b-a7c1-45d7-91bd-451362e096c1/edit-table/2bc08251-4898-42d9-b585-43a11c808a21

edit the sorting of the column and save
the system won't save the setting and prevent the redirection.

kharbutli-meta:
- Fixed table footer width problem.
- Fixed the column highlight problem.
- Cleaned up the prop selectedHeader which is no longer used.
- The update submission issue is a pre-existing problem, will look into it in the near future.
- Layout changes can be discussed for a follow up PR.

mabreuortega (commented):
LGTM.

Question: why the selectedHeader is going away? I think that this still makes sense in the new API

mabreuortega (approved):

PR #965, opened 2022-12-08
fix(frontend): change global colors

denisquesada:
## Description

 A global style change that increase the contrast ratio to 4.8:1.
Change `#C05600` to `#BC4D01` wherever the hex code occurs on the app within CSS.

## Testing

Go to https://d2zdyv8kmq5ffn.cloudfront.net/admin/settings/brandingandstyling/editcolors
Verify that `#C05600` has been replaced by  `#BC4D01` in the **Data visualization second color (optional)** dropdown.

---

### DIFF
--- a/frontend/src/components/Combobox.css
+++ b/frontend/src/components/Combobox.css
@@ -14,9 +14,9 @@
   content: "pla";
 }
 
-.usa-combo-box__list-option[data-value="#c05600"]::after {
-  background-color: #c05600;
-  color: #c05600;
+.usa-combo-box__list-option[data-value="#BC4D01"]::after {
+  background-color: #BC4D01;
+  color: #BC4D01;
   margin: 5px;
   border-radius: 4px;
   content: "pla";
--- a/frontend/src/components/PublishDashboardModal.tsx
+++ b/frontend/src/components/PublishDashboardModal.tsx
@@ -95,7 +95,7 @@ function PublishDashboardModal(props: PathParams) {
                 }
             } catch (err) {
                 setPublishError(
-                    err?.response?.status === 409
+                    (err as any)?.response?.status === 409
                         ? t("PublishDashboardModal.FailToPublishUrlAlreadyExists")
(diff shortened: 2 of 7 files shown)


frontend/src/components/PublishDashboardModal.tsx
```diff
@@ -95,7 +95,7 @@ function PublishDashboardModal(props: PathParams) {
                 }
             } catch (err) {
```
mabreuortega, line 96: err: any should do the same

frontend/src/components/Combobox.css
```diff
@@ -14,9 +14,9 @@
   content: "pla";
 }
 
-.usa-combo-box__list-option[data-value="#c05600"]::after {
-  background-color: #c05600;
-  color: #c05600;
+.usa-combo-box__list-option[data-value="#BC4D01"]::after {
+  background-color: #BC4D01;
+  color: #BC4D01;
```
mabreuortega, line 19: This should be using color from the theme itself, it is good for now but in the future this needs to be converted to SASS and use color("color_name") from the theme palette

frontend/src/services/DatasetParsingService.ts
```diff
@@ -39,7 +39,7 @@ function getFilteredJson(json: Array<any>, hiddenColumns: Set<string>): Array<an
             obj[key] = row[key];
             return obj;
         }, {});
-        if (filteredRow !== {}) {
+        if (Object.keys(filteredRow).length) {
```
mabreuortega, line 42: nitpick: Object.keys(filteredRow).length > 0

mabreuortega (approved):
LGTM, suggestion added

frontend/src/components/PublishDashboardModal.tsx
denisquesada: When adding `catch(err:any)` type scripts throws the error as follows:
```
TS1196: Catch clause variable cannot have a type annotation.

